PR #30, opened 2020-10-19
feature: proxy trezord calls in docker container

szymonlesisz:


docker/compose.yml
```diff
@@ -4,15 +4,24 @@ services:
     container_name: trezor-user-env.unix
     image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env
     network_mode: "host"
+    build:
+      context: ..
+      dockerfile: ./docker/Dockerfile
```
tsusanka, line 9: This is just a leftover right? Because I would still build via `pull` from the registry.

tsusanka (changes requested):
Two tiny things. I can do those myself right away if you want.

tsusanka:
Merged via eb3ad73. Thanks!

PR #32, opened 2020-10-27
controller(nix): workaround udev dependency on darwin

mmilata:


mmilata:
Already merged as part of #33, closing.

PR #33, opened 2020-10-28
Darwin's branch

darwin:
This is my branch, I continue on top of https://github.com/trezor/trezor-user-env/pull/32. It is a continued effort from https://github.com/trezor/trezor-suite/pull/2749.

My goal is to have it working under macOS. I'm also testing this in parallel under WSL2(ubuntu) on Windows 10. Learning Nix in the process.The utlimate goal is to have headless cyperss running both on macOS and Linux hosts.

The original code is pretty much broken for me on macOS and seems to be a bit outdated.

Issues encountered:
1. I had issues downloading firmwares, had a [small yak shaving excursion](https://github.com/trezor/trezor-firmware/pull/1333) into building them myself
2. Even if I had firmwares for correct platform on in proper folder, the python controller server wasn't picking them up - wrong working folder
3. firmware/build.sh is definitely outdated, I can take stab at it
4. run_e2e_tests.sh is also outdated and has no chance to work

I did my own hacky versions of those but I don't know how to move forward. Here is only work I'd like to merge eventually.

### DIFF
--- a/controller/main.py
+++ b/controller/main.py
@@ -2,6 +2,7 @@ import atexit
 import json
 import os
 import binaries
+import argparse
 
 from termcolor import colored
 
@@ -20,8 +21,6 @@ def cleanup():
 atexit.register(cleanup)
 
 PORT = 9001
-
-print("starting websocket server on port: " + str(PORT))
 
 
 # Called for every client connecting (after handshake)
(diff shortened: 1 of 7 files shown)


tsusanka:
This looks good! Could you run `cd controller && nix-shell --run "black ."`? It will format the Python files. And I'll merge.

---

> My goal is to have it working under macOS. I'm also testing this in parallel under WSL2(ubuntu) on Windows 10. Learning Nix in the process.The utlimate goal is to have headless cyperss running both on macOS and Linux hosts.

That'd be great but I honestly felt it might be too much work since we can use docker for that. Did you manage to make it? :)

> I had issues downloading firmwares, had a small yak shaving excursion into building them myself

Yeah it's behind our VPN. We can put them publicly eventually.

> firmware/build.sh is definitely outdated, I can take stab at it

Yes they are, it is a bit of a leftover since the days when I thought we might build that way. But since now the firmware master is fetched directly [from Gitlab](https://github.com/trezor/trezor-user-env/blob/master/firmware/bin/download.sh#L5) on Gitlab I am not sure it makes sense to keep it?

darwin:
> Did you manage to make it? :)

Yes, with `xvfb-run` I was able to run cypress under Linux. It wasn't passing all tests but it was definitely talking to the controller and some tests were passing. I didn't get to running it under macOS, because I didn't have macOS binaries for firmwares. The `./docker_build.sh` script from firmware repo produces only linux versions. And I didn't want to spend time on building it directly on my machine. Next time :)

> I am not sure it makes sense to keep it?

Well, I still think it would have some value in case Gitlab is down or someone wants to build it from sources, e.g. having some changes to test. Of course they could follow all readmes and configure their machine to build the firmware directly on their OS, but I believe a maintained helper script which does all the work would be nice.

darwin:
That said, if we were to provide `firmware/build.sh` it should be using some script from trezor-firmware repo. Which would be more likely to stay maintained.

controller/main.py
```diff
@@ -136,11 +135,36 @@ def message_received(client, server, message):
...
-server.set_fn_new_client(new_client)
-server.set_fn_client_left(client_left)
-server.set_fn_message_received(message_received)
-print("websocket server running ")
-server.run_forever()
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-v", "--verbosity", action="count", default=0)
+    parser.add_argument("--home")
+    parser.add_argument("--enable-proxy", action="store_true")
```
tsusanka, line 142: I think enable-proxy should be on by default. Any reason to disable it btw?

controller/main.py
```diff
@@ -136,11 +135,36 @@ def message_received(client, server, message):
...
-server = WebsocketServer(PORT)
-server.set_fn_new_client(new_client)
-server.set_fn_client_left(client_left)
-server.set_fn_message_received(message_received)
-print("websocket server running ")
-server.run_forever()
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-v", "--verbosity", action="count", default=0)
+    parser.add_argument("--home")
```
tsusanka, line 141: Is `home` is a good name? :thinking. How about `working_dir`?

tsusanka (commented):
Two nits but other than that, great changes! Also no need to run `black` as instructed above, I've checked.

controller/main.py
darwin: Ok, will change.

controller/main.py
darwin: Ok.

controller/main.py
darwin: changed to --work-dir

tsusanka:
Thanks

PR #38, opened 2020-11-27
press-yess press-no instead of decision

mroz22:


tsusanka:
Merged via 1e5f95f7f4acbe3ace77b673b676df11800ed2e6. I have removed the read and confirm mnemonic as discussed.

PR #39, opened 2020-11-27
heavy wip

mroz22:


controller/buttons.py
```diff
@@ -0,0 +1,50 @@
+DISPLAY_WIDTH = 240
```
mroz22, line 1: copy pasta

controller/emulator.py
```diff
@@ -227,3 +242,5 @@ def allow_unsafe():
     time.sleep(SLEEP)
     device.apply_settings(client, safety_checks=1)  # TODO
     client.close()
+
+read_and_confirm_mnemonic()
```
mroz22, line 246: aha, this was here to test locally

mroz22:
Ok, I successfully tested this one locally, so please merge so that we can try it in CI

tsusanka (approved):

PR #51, opened 2021-03-29
tests(suite): shamir pass through backup

wendys-cats:
@tsusanka here you go!

### DIFF
--- a/controller/emulator.py
+++ b/controller/emulator.py
@@ -219,6 +219,85 @@ def read_and_confirm_mnemonic():
     client.press_yes()
     client.close()
 
+def read_and_confirm_mnemonic_shamir():
+    client = DebugLink(get_device().find_debug())
+    client.open()
+    time.sleep(SLEEP)
+
+    client.press_yes()
+
+    # one group
+
+    client.swipe_left()
+    time.sleep(SLEEP)
+
+    client.swipe_left()
+    time.sleep(SLEEP)
+
+    client.swipe_left()
+    time.sleep(SLEEP)
+
+    client.swipe_left()
+    time.sleep(SLEEP)
+
+    client.press_yes()
+    time.sleep(SLEEP)
+
+    # one share
+
+    client.press_yes()
+    time.sleep(SLEEP)
+
+    # set the treshold for one
+
+    client.press_yes()
+    time.sleep(SLEEP)
+    client.press_yes()
+    time.sleep(SLEEP)
+    client.press_yes()
+    time.sleep(SLEEP)
+
+    client.press_yes()
+    time.sleep(SLEEP)
+
+    mnem = client.state().mnemonic_secret.decode("utf-8")
+    mnemonic = mnem.split()
+    time.sleep(SLEEP)
+
+    client.swipe_up()
+    time.sleep(SLEEP)
+
+    client.swipe_up()
+    time.sleep(SLEEP)
+
+    client.swipe_up()
+    time.sleep(SLEEP)
+
+    client.swipe_up()
+    time.sleep(SLEEP)
+
+    client.swipe_up()
+    time.sleep(SLEEP)
+
+    client.press_yes()
+    time.sleep(SLEEP)
+
+    index = client.read_reset_word_pos()
+    client.input(mnemonic[index])
+    time.sleep(SLEEP)
+
+    index = client.read_reset_word_pos()
+    client.input(mnemonic[index])
+    time.sleep(SLEEP)
+
+    index = client.read_reset_word_pos()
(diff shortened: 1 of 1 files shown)


tsusanka:
@grdddj in this PR @wendys-cats added a "walkthrough" Shamir Backup. We have a command that clicks through normal backup but we did not have one for Shamir. Could you please review it and confirm it works?

grdddj:
Sure, already taking a look at that, but it seems I will need some help with this. So far I am not even able to run the non-shamir `read_and_confirm_mnemonic()` function. It seems that I am missing some prerequisite to run this command. Errors I get are below - 1 happens when trying that for the first time in session, and 2 follows all the time after it.

Steps I am following:
- run the bridge 2.0.31
- run the emulator 2.4.0 (but tried even older ones)
- putting `{"type": "emulator-read-and-confirm-mnemonic", "id": 123}` into the JSON console to trigger this function
- (also, some combination of removing the `var/tmp/trezor.flash` profile and trying to setup the device via [wallet](https://wallet.trezor.io) or [suite](https://suite.trezor.io/web/))

@wendys-cats could you please advise what is the correct setup before running these functions?

NOTE: I am running these already on our latest `incoming` branch, I tried on `wendys-cats:shamir` first, but as it is based on quite old version of `master`, it has very little debugging possibilities.

1
image

2
image

grdddj:
I managed to get the emulator into the state of being able to run those two functions `read_and_confirm_mnemonic()` and `read_and_confirm_mnemonic_shamir()`, and the first one is fine.

The second (Shamir) one however has issue, as at the beginning of setup (screen below) we want to decrease the number of shares to one, but the `swipe_left()` function does not do that. **We would like to click the minus button**

@tsusanka By any chance, don't you know (or don't you know somebody who knows) how to programmatically click the minus button? I am looking through the `trezorlib` code right now.

image

tsusanka:
@matejcik will know :)

grdddj:
Thanks @matejcik for help, the `client.click((x,y))` is doing the trick (after estimating the coordinates of the button).

Another issue right now is **how to retrieve the correct word at a specific location** when doing the "quiz" (screen below)

In classic non-shamir backup, we are getting the seed words by code below, as the `mnemonic_secret` has the value of  `b'melody average toast dish expand thrive limb tower divide flee history volume'`
https://github.com/trezor/trezor-user-env/blob/b0b19668a00b4a36db59a9436f56654a6988978e/src/emulator.py#L234

However, in case of shamir-backup, the `mnemonic_secret` has value of `b'z\xc4\xf8\x02\\\x16g\\\t\xc7@\x01\xb4\xc7\xf8~'`, which cannot be decoded into string (`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc4 in position 1: invalid continuation byte`).

Apart from the `mnemonic_secret`, the `client.state()` object does not contain any useful data, apart from `reset_entropy`.

I am investigating this, but would appreciate any help.

(**Possible dirty approach**: when swiping through the 20 words, we can gradually save them, as they are being available in `layout_lines` property of the  `client.state()` object)

image

matejcik:
> (**Possible dirty approach**: when swiping through the 20 words, we can gradually save them, as they are being available in `layout_lines` property of the `client.state()` object)

basically this but instead of `layout_lines` use `debuglink.read_reset_words()`

see how the test suite does it: https://github.com/trezor/trezor-firmware/blob/master/tests/common.py#L178-L209

matejcik:
for the record this is what you'd need to do if you wanted to rely on `layout_lines`:
https://github.com/trezor/trezor-firmware/blob/master/tests/click_tests/reset.py#L28-L69

grdddj:
Thanks, it is working fine!

@tsusanka How will we do with this PR? As I do cannot push into @wendys-cats's fork branch, I have created a local `shamir` branch, so we can create a new PR (it contains that one commit from @wendys-cats). However, I would not like to just close/dismiss this PR if it was not necessary. Any suggestions?

tsusanka:
Let's create a new PR. @wendys-cats' contribution will stay if you just cherry-pick her commit so we won't take the credit from her :).

grdddj:
Agreed. In current [shamir branch](https://github.com/trezor/trezor-user-env/commits/shamir) the commit is shows both @wendys-cats  and me, as I had to resolve the conflicts (we moved the file that was being modified).

Can it be done better?

I will be finishing my local changes, push it on `shamir` branch, rebase on `incoming` and create a new PR.

tsusanka:
It can't. You are co-contributors. But it says exactly what you did `wendys-cats authored and grdddj committed` so I think we are all good :).

PR #57, opened 2021-06-24
Grdddj/code quality

grdddj:
Making the code more pythonic, mypy testing, formatting by black, deleting suite directory and html suite part


grdddj:
This is already outdated, all the still relevant changes were migrated into the **servers** branch

tsusanka:
So let's close this one and I review #58 instead?

grdddj:
> So let's close this one and I review #58 instead?

Yes, agreed, this one does not contain anything that is not in #58 (and also is based on master which is outdated)

Issue #48, opened 2021-01-26
Merge shell.nix and controller/shell.nix


tsusanka:
There is little reason to have it separated and it confused people. Let's merge it.

tsusanka:
Done in `incoming`, soon in `master` :tada:.

PR #62, opened 2021-07-01
feat(nix-shell): merging both nix-shells into one in root folder

grdddj:
Addressing issue #48 - combining both nix-shells into one in the root folder, to make it less confusing


grdddj:
Fixup commit c6e0bf4:
- reflecting the change in README

shell.nix
```diff
@@ -1,13 +1,31 @@
...
-    wget
-  ];
-}
+# the last successful build of nixos-20.09 (stable) as of 2020-10-11
+with import
+  (builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/0b8799ecaaf0dc6b4c11583a3c96ca5b40fcfdfb.tar.gz";
+    sha256 = "11m4aig6cv0zi3gbq2xn9by29cfvnsxgzf9qsvz67qr0yq29ryyz";
+  })
+{ };
```
tsusanka, line 9: @mmilata if we wish to update to latest build of nixos-21.05 (stable) we'll modify the commit to the latest one in `release-21.05` [branch](https://github.com/NixOS/nixpkgs/tree/release-21.05)? Or is it a different one?

mmilata: Preferably in `nixos-21.05`, `release-21.05` is the same but before passing CI. I replace the sha256 with 52 zeroes, then replace it again with the one `nix-shell` prints in the error message. Or you can use `nix-prefetch-url` to get the hash.

tsusanka: Good trick :). So I got commit `e9148dc1c30e02aae80cc52f68ceb37b772066f3` and hash `1xs5all93r3fg4ld13s9lmzri0bgq25d9dydb08caqa7pc10f5ki`. So let's try that?

grdddj: Yes, seems to work fine, I am about to commit these changes. Thanks for the help

grdddj:
Commit 7edfdd5:
- adding latest version of Nixos build into the shell.nix, version 21.05

tsusanka (approved):
Great. Let's squash the fixup commit and let's merge :).

tsusanka (approved):

tsusanka:
So the job seems to pass! https://gitlab.com/satoshilabs/trezor/trezor-suite/-/jobs/1399269501

So we have successfully added a commit into the `incoming` branch that does not seem to break anything! Or I messed up the "testing" but hopefully not :).

grdddj:
> So the job seems to pass! https://gitlab.com/satoshilabs/trezor/trezor-suite/-/jobs/1399269501
> 
> So we have successfully added a commit into the `incoming` branch that does not seem to break anything! Or I messed up the "testing" but hopefully not :).

Good job! Nice that we have this testing pipeline to be more confident.

PR #63, opened 2021-07-02
Grdddj/make style checks

grdddj:
Including basic style checks in Makefile and connecting it with CI

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,6 +2,11 @@ variables:
   # No need for TLS because everything runs on the same machine
   DOCKER_TLS_CERTDIR: ""
 
+style prebuild:
+  stage: .pre
+  script:
+    - nix-shell --run "make style_check"
+
 .build: &build
   image: docker
   variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,39 +2,10 @@ variables:
   # No need for TLS because everything runs on the same machine
   DOCKER_TLS_CERTDIR: ""
 
-style prebuild:
-  stage: .pre
-  script:
-    - nix-shell --run "make style_check"
+stages:
+  - prebuild
+  - build
 
-.build: &build
-  image: docker
-  variables:
-    CONTAINER_NAME: "$CI_REGISTRY/satoshilabs/trezor/trezor-user-env/trezor-user-env"
-  services:
-    - docker:dind
-  before_script:
-    - docker login $CI_REGISTRY -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD
-
-environment:
-  <<: *build
-  only:
-    - schedules  # nightly build
-    - master
-    - /^run\//
-  script:
-    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:latest -f docker/Dockerfile .
-    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
-    - docker push $CONTAINER_NAME:latest
-
-environment test:
-  <<: *build
-  except:
-    - schedules
-    - master
-    - /^run\//
-  script:
-    # note that this way "test" tag gets overwritten everytime
-    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:test -f docker/Dockerfile .
-    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
-    - docker push $CONTAINER_NAME:test
+include:
+  - ci/prebuild.yml
+  - ci/build.yml
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -1,0 +1,33 @@
+.build: &build
+  image: docker
+  variables:
+    CONTAINER_NAME: "$CI_REGISTRY/satoshilabs/trezor/trezor-user-env/trezor-user-env"
+  services:
+    - docker:dind
+  before_script:
+    - docker login $CI_REGISTRY -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD
+
+build and push image:
+  stage: build
+  <<: *build
+  only:
+    - schedules  # nightly build
+    - master
+    - /^run\//
+  script:
+    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:latest -f docker/Dockerfile .
+    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
+    - docker push $CONTAINER_NAME:latest
+
+build and push test image:
+  stage: build
+  <<: *build
+  except:
+    - schedules
+    - master
+    - /^run\//
+  script:
+    # note that this way "test" tag gets overwritten everytime
+    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:test -f docker/Dockerfile .
+    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
+    - docker push $CONTAINER_NAME:test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+style_check:
+	isort --version | awk '/VERSION/{print $$2}'
+	black --version
+	@echo [ISORT]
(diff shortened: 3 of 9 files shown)


grdddj:
Couple issues encountered:
- all the commits from the merge-nix-shells branch are here as well (my fault, sorry for that, I should have really started only on fresh master)
- the CI checks are currently failing, seems like the "prebuild" stage does not exist:
image
(possible solution seems to change the stage to ".pre")
   - the CI right now **is** supposed to fail, as the style_checks are failing (and I did not want to commit all those formatting changes here - in servers branch, they are already fixed)

grdddj:
Fixup commit bf5081e:
- fixing the issue with unexisting stage, was changed to *.pre*
- now the check is failing most probably because of bad styling, but it somehow did not show the exact reason (and the line 20 of the log is missing, see below)

image

grdddj:
@tsusanka Next considerations about CI:
- adding Poetry with locked library versions?
- adding mypy and flake8 checks into some command like *make code_check*?

Both seem easy to do (almost a copy-paste from firmware repo), but they may be an overkill for such a small project right now

tsusanka:
The stage did not exist because we have not defined it (I have probably confused you a little bit when we spoke), see [firmware](https://github.com/trezor/trezor-firmware/blob/master/.gitlab-ci.yml#L22-L28).

The trick with `.pre` probably works because it is some kind of default from Gitlab? Not sure.

Anyway, let's create two stages:
```
stages:
  - prebuild
  - build
```

And [assign](https://github.com/trezor/trezor-firmware/blob/master/ci/build.yml#L10) `prebuild` to the style check and the other two jobs to the `build`.

vdovhanych:
I will commit changes to the GitLab ci configuration file and set up a better CI folder structure right away, so we have it from the start. If that is ok with you guys @tsusanka @grdddj

vdovhanych:
> Fixup commit [bf5081e](https://github.com/trezor/trezor-user-env/commit/bf5081e6f1682936cc335940db92e5492c20dec1):
> 
> * fixing the issue with unexisting stage, was changed to _.pre_
> * now the check is failing most probably because of bad styling, but it somehow did not show the exact reason (and the line 20 of the log is missing, see below)
> 
> image

this has failed because as I stated above you were trying to run nix-shell on who knows what by default and it didn't run at all.

grdddj:
> I will commit changes here to the GitLab ci configuration file and set up a better CI folder structure right away, so we have it from the start. If that is ok with you guys @tsusanka @grdddj

Those improvements in your branch look good, thank you for them! Feel free to include them in this branch (at least I am not against - and if it is even possible to commit in other people's Pull-Requests).

> As I stated above, this has failed because you were trying to run nix-shell on who knows what by default, and it didn't run at all. Because there was no nix-shell present in the env, it was running in.

Good catch with that base image, we cannot just run it from thin air. Theoretically this `nixos/nix` is enough for everything, it should not need any other additions, as running `nix-shell` should install all the dependencies itself from the *shell.nix* file, where we define them. (Please correct me here, I am just beginning with NixOS)

The job from your last commit - https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/1397205290 - is failing on the style-check, exactly as expected, this branch does not have any styling yet (can be run with `make style`).

Also, with @tsusanka we have agreed that we should be cautious with not breaking anything in the Suite CI, so instead of `master` branch we will be merging into `incoming` branch first, test it against Suite and only then continue into `master`, if successful.

tsusanka:
Please let's:
1. Rebase on top of new `incoming`, the shell.nix changes are there.
2. Add commits from @vdovhanych about CI.
Feel free to squash and force push in case it becomes easier.

vdovhanych:
Hi all, I went ahead and pushed the commit with all the changes to the ci right to this branch, so it should be easier to just rebase with incoming.

grdddj:
> Hi all, I went ahead and pushed the commit with all the changes to the ci right to this branch, so it should be easier to just rebase with incoming.

Hello, thanks for your commit, I will try to rebase and force-push now

grdddj:
After the rebase and force-push, the negative CI result is again a positive one as it fails as expected on the styling check

.gitlab-ci.yml
```diff
@@ -2,34 +2,10 @@ variables:
...
-    - master
-    - /^run\//
-  script:
-    # note that this way "test" tag gets overwritten everytime
-    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:test -f docker/Dockerfile .
-    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
-    - docker push $CONTAINER_NAME:test
+include:
+  - ci/prebuild.yml
+  - ci/build.yml
```
tsusanka, line 11: Mostly an OCD thing and probably from @vdovhanych but please let's have a newline at the end in each file.

ci/build.yml
```diff
@@ -0,0 +1,33 @@
...
+  only:
+    - schedules  # nightly build
+    - master
+    - /^run\//
+  script:
+    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:latest -f docker/Dockerfile .
+    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
+    - docker push $CONTAINER_NAME:latest
+
+build and push test iamge:
```
tsusanka, line 22: Typo (most likely from me :) ) `iamge` -> `image`.

ci/build.yml
```diff
@@ -0,0 +1,33 @@
...
+  <<: *build
+  except:
+    - schedules
+    - master
+    - /^run\//
+  script:
+    # note that this way "test" tag gets overwritten everytime
+    - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:test -f docker/Dockerfile .
+    - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
+    - docker push $CONTAINER_NAME:test
```
tsusanka, line 33: New line here as well please :).

tsusanka (commented):
Few comments. Let's now also add a commit which runs the `make style` and `make isort` so the CI passes.

grdddj:
FIxup commit 53e1ffa:
- fixing the small inconsistencies from the comment above

Commit fixing the style will follow

grdddj:
Commit c0ca7e1:
- running `make style` to make sure the `make style_check` test in CI is alright

tsusanka:
Nix should make sure we have the same versions as our nix-shell is pinned to [particular commit](https://github.com/trezor/trezor-user-env/commit/22a75db064d1103b0d122a5691267a76175bfce4) therefore it is reproducible always to the same software.

My guess is that you have not exited from the nix-shell and therefore you are still on inside the older `shell.nix`. Try to exit and enter again on this branch.

tsusanka:
Unrelated to that: I would consider removing [this](https://github.com/trezor/trezor-user-env/pull/63/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R1) and simply run `black .` and `isort .` directly. I think the `find` was in trezor-firmware due to the fact that it is a monorepo and it was doing some nasty things.

grdddj:
> Nix should make sure we have the same versions as our nix-shell is pinned to [particular commit](https://github.com/trezor/trezor-user-env/commit/22a75db064d1103b0d122a5691267a76175bfce4) therefore the system should be completely reproducible always to the same software.
> 
> My guess is that you have not exited from the nix-shell and therefore you are still on inside the older `shell.nix`. Try to exit and enter again on this branch.

You were right, after restarting the nix-shell, I got the same versions as in the CI. Fixed (hopefully) in fixup commit 51201d0

grdddj:
> Unrelated to that: I would consider removing [this](https://github.com/trezor/trezor-user-env/pull/63/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R1) and simply run `black .` and `isort .` directly. I think the `find` was in trezor-firmware due to the fact that it is a monorepo and it was doing some nasty things.

Quite a good simplification, I like it. Included in 9427f5e

tsusanka (approved):
Looks great! Let's squash and force push.

tsusanka (approved):

PR #66, opened 2021-07-05
Grdddj/bridge binaries update

grdddj:
Experimental branch to check if bridge binaries update on NixOS will not break anything


tsusanka:
As per our [Dockerfile](https://github.com/trezor/trezor-user-env/blob/master/docker/Dockerfile#L26) the `download.sh` is run for Bridge anyway, so this indeed should not break anything. Let's see :).

PR #67, opened 2021-07-06
feat(ci): code checks with mypy and flake8

grdddj:
Including code checks into CI and making an easy `make check` command to check everything locally.

NOTE: in the #58 PR, there is a `tests` folder with mypy testing file, which should be then deleted, as this is much better approach.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 88
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .idea
+.vscode
 suite/repo
 firmware/repo
 trezord-go/repo
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+check: style_check code_check
+
 style_check:
 	isort --version | awk '/VERSION/{print $$2}'
 	black --version
@@ -11,3 +13,9 @@ style:
 	@isort .
 	@echo [BLACK]
 	@black .
+
+code_check:
+	@echo [MYPY]
+	@mypy .
+	@echo [FLAKE8]
+	@flake8 .
--- a/ci/prebuild.yml
+++ b/ci/prebuild.yml
@@ -1,5 +1,11 @@
-full code check:
-  image: nixos/nix:latest
+image: nixos/nix:latest
+
+code check prebuild:
   stage: prebuild
   script:
-    - nix-shell --run "make check"
+    - nix-shell --run "make code_check"
+
+style check prebuild:
+  stage: prebuild
+  script:
+    - nix-shell --run "make style_check"
--- a/ci/prebuild.yml
+++ b/ci/prebuild.yml
@@ -1,5 +1,5 @@
-style check:
+full code check:
   image: nixos/nix:latest
   stage: prebuild
   script:
-    - nix-shell --run "make style_check"
+    - nix-shell --run "make check"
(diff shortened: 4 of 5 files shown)


grdddj:
The current CI is failing because of mypy issues, which are already fixed in #58.

Before merging this, we can make a fixup commit here to make it succeed to verify it.

grdddj:
Fixup commit da4eddf:
- Adding a newline at the end of `setup.cfg` file

tsusanka (approved):
Great :+1:. Since we are merging into `incoming` I do not mind merging a failing PR, rebase #58 on top of it and make sure it passes there. To simplify things. Sounds okay?

grdddj:
> Great +1. Since we are merging into `incoming` I do not mind merging a failing PR, rebase #58 on top of it and make sure it passes there. To simplify things. Sounds okay?

I would agree. Let me squash these two commits into one and force-push here

vdovhanych (approved):
Looks good

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 120
```
tsusanka, line 2: Btw I am not sure this is needed as `black` formats lines to max 88 (or smth like that).

grdddj: Good point! However, when not specified, flake has the upper limit of 79, so it would still scream

tsusanka: Ah, funny :). But let's put it to 88 so we are inline with black? It is indeed 88 https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

grdddj: Right, thanks for the source, I will change it.

They even say `If you’re using Flake8, you can bump max-line-length to 88 and mostly forget about it.`

tsusanka: Great :+1:

tsusanka:
I have realized one more nit-pick, sorry. It is slightly better to actually have two jobs in the CI, one for style, one for mypy. Only for the simple reason that you see right away what went wrong. So if you do not mind I would split the `full code check` into `style` and `mypy` jobs.

If you find the full check handy for local development feel free to keep it in the Makefile.

grdddj:
> I have realized one more nit-pick, sorry. It is slightly better to actually have two jobs in the CI, one for style, one for mypy. Only for the simple reason that you see right away what went wrong. So if you do not mind I would split the `full code check` into `style` and `mypy` jobs.
> 
> If you find the full check handy for local development feel free to keep it in the Makefile.

No issue, I can split it. The code_check should probably go first, as it is more important than the style_check.

And yeah, I really like the ability of running simple `make check` locally, so will keep it there

grdddj:
In 9e5e0f6 I tried an experiment with refactoring common information out of the jobs - inspired by [firmware](https://github.com/grdddj/trezor-firmware/blob/master/ci/prebuild.yml#L1)

Turned out the stage needs to belong to every script.

Fixed in e97e2eb

vdovhanych (approved):
Also, fine with me, I have some nitpicks that there are no newlines at the end of the yml file, but that is no big deal, just my OCD (and yeah, I think I made a mistake when I first committed it here) will add `.editorconfig` file, so we don't have to think about this.

tsusanka (approved):
Great, feel free to squash and let's merge.

tsusanka:
> No issue, I can split it. The code_check should probably go first, as it is more important than the style_check.

Btw all jobs inside one stage run at the same time (well up to the number of allowed parallel jobs which is 8 ATM I believe).

PR #68, opened 2021-07-07
feat: add editorconfig file to the project

vdovhanych:
Added the editorconfig file, so we do not have to think about those small mistakes, no newlines at the end and so on.


vdovhanych:
sorry, wrong branch :(

grdddj:
> sorry, wrong branch :(

If you just meant to merge it against `incoming` and accidentally chose `master`, it can be easily changed even on existing PR with the `Edit` button on the right:

image

PR #69, opened 2021-07-07
feat: add editorconfig file to the project

vdovhanych:
Added the editorconfig file, so we do not have to think about those small mistakes, no newlines at the end and so on.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# EditorConfig configuration for trezor-user-env
+# https://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file, utf-8 charset
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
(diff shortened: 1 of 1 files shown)


grdddj:
Hello, thanks for this, works fine for me :)

When I saved the new `.editorconfig`  file itself, it added a newline at the end as expected (I imagine it was the purpose of the newline at the end missing :D)

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+# Ignore diffs/patches
+[*.{diff,patch}]
+end_of_line = unset
+insert_final_newline = unset
+trim_trailing_whitespace = unset
```
tsusanka, line 18: Yes, comically you have forgotten to add a newline at the end of your configuration file which enforces newlines :laughing:

vdovhanych: yep, because when i did it it removed the file and didn't fix it 🤦  Pushed update

Issue #65, opened 2021-07-05
Bake git revision into the Docker image


tsusanka:
It would be great to somehow bake git revision (git commit hash) into the docker image. We could print it then in CI and know on which trezor-user-env version we are.

grdddj:
Theoretically we can have a script in Dockerfile that will create something like `version.txt` file inside the image.

This could include the number of commits in the current branch (as a relative identifier of version - the higher amount of commits, the higher the version, plus a branch name) and also detailed information about last N (3?) commits (as running `git log -3`).

It can be then printed as the last pre-CMD step of the Dockerfile, so whenever the container will run in CI tests, this info will be shown.

tsusanka:
Yeah something like that, but I am wondering there might be some best practice here. @vdovhanych maybe you could ask Ondra on some occasion?

PR #71, opened 2021-07-10
feat(docker): showing version details when running docker container

grdddj:
Trying to address issue #65
- when running a docker container, it shows multiple git details - the branch, commit amount (aka version) and the details of last commit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ src/binaries/firmware/bin/*
 emulator.img
 .vscode
 .mypy_cache
+docker/version.txt
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,6 +25,8 @@ RUN ls /trezor-user-env
 RUN nix-shell --run "./src/binaries/firmware/bin/download.sh"
 RUN nix-shell --run "./src/binaries/trezord-go/bin/download.sh"
 
+RUN nix-shell --run "./docker/create_version_file.sh"
+
 RUN rm -rf /var/tmp/trezor.flash
 
 RUN nix-shell --run "echo deps pre-installed"
--- a/docker/create_version_file.sh
+++ b/docker/create_version_file.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env sh
+
+DIR=$(dirname "$0")
+cd "${DIR}"
+
+version_file="version.txt"
+
+# Including the branch and version information plus details from last commit(s)
+commit_amount=$(git rev-list HEAD --count)
+branch=$(git branch --show-current)
+last_commit_info=$(git log -1 --stat)
+echo "Trezor-user-env information" > ${version_file}
+echo "Version (commit amount): ${commit_amount}" >> ${version_file}
+echo "Branch: ${branch}" >> ${version_file}
+echo "" >> ${version_file}
+echo "${last_commit_info}" >> ${version_file}
+echo "********************************************************" >> ${version_file}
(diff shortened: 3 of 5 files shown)


grdddj:
It is probably off-topic in this PR, but when trying to run the docker container (built locally to test this version functionality), I received an error in `run.sh`, in `nix-shell --run "trezorctl --version"` command - could not find version.

Has it ever been encountered?

NOTE: For my purposes, I have disabled all the `download.sh` lines in Dockerfile, to have it built faster - theoretically that may be the culprit, but these should have nothing to do with `trezorlib`, it is defined only in `shell.nix`

image

grdddj:
UPDATE on error above:
- the whole trezorlib is working fine apart from this `--version` command
- even locally the same command - `nix-shell --run "trezorctl --version"` throws the same error
- (this `--version` is a default `click` functionality)
- when running "our" command for version in trezorlib - `nix-shell --run "trezorctl version"` (without the "--"), it is working fine (comparison below)

I know that I am hijacking this PR, but since I already made one comment about it here, am following with this here as well. Is it worth creating a separate issue for this?

image

tsusanka:
Aaah and have we ever even supported `trezorctl --version`? :thinking: Maybe it was just `trezorctl version` all the time and it was incorrect all the time here in trezor-user-env. Could you please create an issue in trezor-firmware? Matejcik will advise there and it is indeed off-topic here.

tsusanka:
And in the meantime please remove the `--` here in this PR.

tsusanka (commented):
This is great :clap: :clap:. Please:
- Let's move `create_version_file.sh` into the `docker`
- Same for the `version.txt` although that does not matter that much
- Add `docker/version.txt` to `.gitignore` so we do not create a possible conflict in future
- Remove the `--` in `trezortl --version`
- Rebase on top of the new `incoming`

Feel free to force-push directly, this is a tiny PR so we'll manage :).

grdddj:
> Aaah and have we ever even supported `trezorctl --version`?

Yes, under "normal" circumstances, this command seems to work out of the box from the `click` library used for creating the trezorctl CLI.

Even when I simply get out of the nix-shell and run it with my laptop NixOS, it is fine:
image

Therefore it is probable that there is some incompatibility of `click` library installed by the new shell.nix here (through the `trezor` package)

grdddj:
@tsusanka Meanwhile I am creating the issue in firmware repo, can you please try to run this command as well on your machine (in root folder in this branch), so we know it is not connected only to my PC?

`nix-shell --run "trezorctl --version"`

(I tried to run it on `master` branch (as `nix-shell controller/shell.nix  --run "trezorctl --version"`), and it is fine there - there is older shell.nix file)

tsusanka:
Yes I am getting the same error on 0abb0ffbcdb62b3038e3d446eca27c5b805843b1.

grdddj:
> Yes I am getting the same error on [0abb0ff](https://github.com/trezor/trezor-user-env/commit/0abb0ffbcdb62b3038e3d446eca27c5b805843b1).

Thanks for verifying, on `incoming` it is the same, so with a high probability it is caused by the NixOS update we did.

As mentioned in the [firmware issue](https://github.com/trezor/trezor-firmware/issues/1702), it is not a blocker for us - as we run this command only for informational purposes, and the error does not cause any issue with our main script. However it may be an issue for someone else.

grdddj:
Commit 2cc65c4:
- incorporating the comments above
  - moving the version-file-generating script
  -  changing the _version_ command for trezorctl and reflecting the new output (does not show the _trezorctl_ in any way, is just a version number now - therefore prepending it with a description text)

tsusanka (approved):

tsusanka:
Great! You are moving so fast I'll have to start about new issues :)).

Issue #72, opened 2021-07-10
`incoming` branch fails with Suite


tsusanka:
After merging #58 I have restarted Suite e2e job which [runs against](https://gitlab.com/satoshilabs/trezor/trezor-suite/-/commit/3fce6f8d48197d12d1aa279eaaef703a33bfb5ed) the incoming docker image and [it fails](https://gitlab.com/satoshilabs/trezor/trezor-suite/-/jobs/1414462421):

```
CypressError: `cy.task('stopBridge')` failed with the following error:
> websocket_error_message: RuntimeError - Bridge is not running, cannot be stopped
Because this error occurred during a `before each` hook we are skipping all of the remaining tests.
```

So this is mostly likely caused by the fact that Bridge was not started properly.

What is a bit unpleasant is that the CI job does not print trezor-user-env log. @vdovhanych can you get that for us or do we need to run it manually? @grdddj could you please coordinate with @vdovhanych?

Also I have created an issue https://github.com/satoshilabs/devops/issues/37 to have the log as an artifact (@grdddj you need to accept an invite to see it).

grdddj:
I tried to identify the place that causes the issue, and it seems we are trying to stop the bridge and emulator before every test as a cleanup (even if they do not run)

![image](https://user-images.githubusercontent.com/42543243/125190959-42b6e980-e240-11eb-9897-85ad2eccf2fa.png)

We have actually implemented the `cleanup` argument in our code (`bridge.py`), which when True, will not raise that RuntimeError

![image](https://user-images.githubusercontent.com/42543243/125191010-86115800-e240-11eb-8c4f-33ea6ae5739b.png)

Possible solutions (starting with the easiest one):
- just removing the RuntimeError in our code and add a log statement there (like `WARNING: bridge is stopped but is not running`)
- implementing new function in both suite and trezor-user-env, something like `make_sure_bridge(emulator)_is_stopped`, that can be run before these tests

tsusanka:
I think stopping Bridge even if it is not running is a valid call. So I would do your first suggestion.

tsusanka:
Let's try #75 because I think it is the right thing to do anyway but https://github.com/trezor/trezor-user-env/issues/53#issuecomment-877821922 made me think this might not be the reason the CI fails. The Bridge should be running when running `stop` I guess so we should investigate if the Bridge was indeed spawned.

grdddj:
> I guess so we should investigate if the Bridge was indeed spawned

[...]

PR #75, opened 2021-07-11
Grdddj/ci failing runtime errors

grdddj:
Trying to address issue #72
- do not throw the RuntimeError when stopping the bridge and proxy even if they do not run already (replaced with a warning log)
- RuntimeError still left in both `start()` function, can be also removed if causing issues


grdddj:
Commit 6ecc1e9:
- unifying all the stop() functions - adding the same logging into `emulator` as already in `bridge` and `bridge_proxy`

src/bridge.py
```diff
@@ -47,7 +47,7 @@ def start(version: str, proxy: bool = False) -> None:
     #   not to still think the bridge is running
     if proc is not None and not is_running():
         log("Bridge was probably killed by user manually, resetting local state")
-        stop(cleanup=True, proxy=proxy)
+        stop(proxy=proxy)
 
     if proc is not None:
         raise RuntimeError("Bridge is already running, not spawning a new one")
```
tsusanka, line 53: Should we do the same with `start()`?

grdddj: Yes, we can remove raising this error - but then we need to agree what else to do in case bridge is being started, but it is already running. Most certainly two bridges cannot run at the same time, so we would either ignore the second one or kill the first one.

What is my worry, that without these RuntimeErrors, which communicate the situation to the client, the app will not be as responsive as possible (telling users they probably did something wrong - spawned a bridge two times etc.).

The biggest mystery right now is **Why does raising the RuntimeError in stop() kill the CI, when there is try-except in controller?**.

I would actually prefer to keep these RuntimeErrors, but we need to be really sure they do not cause issues in CI :)

vdovhanych: does it produces an error with a 1? that is probably why it fails if it does.

grdddj: I think I know now - the Suite Cypress tests are processing the response JSON from the websocket, and if there is `success: False`, they will fail the tests - more in #72

tsusanka: Great find :+1:. Ok let's not solve something that does not pose a problem for us at the moment :). Let's keep the Rrror for `start`. Resolving.

tsusanka (approved):

PR #76, opened 2021-07-12
feat(controller): relaying all the details and warnings to client (POC)

grdddj:
Made a quick proof of concept for how to let clients well informed but still satisfy the Cypress tests
- each function called by controller will return three things to it - `response`, `details` and `warnings`
- these things will be relayed to the client, with `success: True` key, as long as there will not be any "real" error

### DIFF
--- a/src/controller.py
+++ b/src/controller.py
@@ -82,8 +82,12 @@ async def handler(websocket, path) -> None:
             elif command == "emulator-start":
                 version = request.get("version") or binaries.FIRMWARES["TT"][0]
                 wipe = request.get("wipe") or False
-                emulator.start(version, wipe)
-                response = {"response": f"Emulator version {version} started"}
+                result = emulator.start(version, wipe)
+                response = {
+                    "response": result.get("response", ""),
+                    "details": result.get("details", []),
+                    "warnings": result.get("warnings", []),
+                }
             elif command == "emulator-stop":
                 emulator.stop()
                 response = {"response": "Emulator stopped"}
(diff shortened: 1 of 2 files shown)


tsusanka:
I am not sure I like it :confused:. It feels we are bloating the API redundantly. I would just leave it as it is - if trezor-user-env user needs more info they simply open and read the log. Don't you think?

grdddj:
> I am not sure I like it confused. It feels we are bloating the API redundantly. I would just leave it as it is - if trezor-user-env user needs more info they simply open and read the log. Don't you think?

I completely agree that it is making the code more messy and complicated, when we would be relaying everything. However, it could then really show what is happening inside, and can therefore uncover some inner issues that would otherwise be unnoticed.

Seems that I got focused more on the connection with HTML page, than on the Cypress CI tests (and you are right, human users can just look into the terminal).

A question to resolve (or reject :D) this confusion - **what is more important for us, the user-side of HTML, or the CI tests?** (after having a thought, the CI is much more crucial, but that does not mean we cannot do both well)

tsusanka:
Yeah I would say CI. Let's ask also here our Cypress guru @mroz22: would you like to have this PR merged? It adds more info the response json (details and warnings) but as mentioned it seems a bit redundant and kind of "dyi" to me.

grdddj:
What is also an easy option to do is have a boolean argument in the request (like `send_details`), defaulting to `False`, so that the HTML site can opt for it, but the Cypress tests will not be affected (and can add this any time)

grdddj:
I have just rebased the PR branch with `incoming`, so it really shows only the relevant stuff

tsusanka:
Upon discussion we have agreed to postpone this until we see clear benefits of this. Let's reopen once we have a use-case which this would help.

PR #78, opened 2021-07-14
Grdddj/controller code quality

grdddj:
Creating a class that is handling the whole websocket request-response traffic (goes well together with #76)
Advantages:
- it is more readable and understandable
- it can be easily modified and extended (for example sending different responses to different clients, based on the request data - if they ask for more details etc.)

### DIFF
--- a/src/controller.py
+++ b/src/controller.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import traceback
-from typing import Any, Dict
+from copy import deepcopy
 
 import websockets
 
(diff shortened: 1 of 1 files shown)


grdddj:
Fixup commit b3010fe:
- not putting the websocket object into the class (so that it has only one responsibility - translating request into response)
- getting request in the main code and only putting that to the object as an argument

tsusanka (approved):
Nice, LGTM. Squash and merge?

grdddj:
> Nice, LGTM. Squash and merge?

Agreed, doing the squash

PR #79, opened 2021-07-16
chore(logs): creating separate folder for logs

grdddj:
Separate folder for logs, so they can be easily extracted from the container in CI jobs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@ emulator.img
 .vscode
 .mypy_cache
 docker/version.txt
-debugging.log
+logs/debugging.log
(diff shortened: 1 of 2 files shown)


tsusanka (commented):
Not that it matter much but we could simply put it into the docker dir? But it is true it will break when we mess around with it, so let's keep it like this. We should also change the `.gitignore`?

grdddj:
> Not that it matter much but we could simply put it into the docker dir? But it is true it will break when we mess around with it, so let's keep it like this. We should also change the `.gitignore`?

Right, I forgot the `.gitignore`, let me change and force-push that.

PR #80, opened 2021-07-16
Grdddj/suite ci issues

grdddj:
Fixing the issue with failing Suite integration tests (#72)
- the `bridge_proxy` server was previously spawned in a different thread, but even after shutting it down it was still blocking the port, and therefore it was impossible to spawn it again after shutting it down
- solving it by spawning it in a different process, taking the same approach as with `bridge`

### DIFF
--- a/src/bridge.py
+++ b/src/bridge.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 import os
 import signal
+import socket
+import time
 from subprocess import Popen
 
 import requests
@@ -8,6 +10,9 @@ from requests.exceptions import ConnectionError
 
 import bridge_proxy
 import helpers
+from bridge_proxy import PORT as BRIDGE_PROXY_PORT
+
+BRIDGE_PORT = 21325
 
 # TODO: consider creating a class from this module to avoid these globals
 proc = None
@@ -38,7 +43,20 @@ def get_status() -> dict:
     return {"is_running": is_running(), "version": version_running}
 
 
+def is_port_in_use(port: int) -> bool:
+    """Checks if a certain port is listening on localhost"""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        return s.connect_ex(("0.0.0.0", port)) == 0
+
+
+def check_bridge_and_proxy_status() -> None:
+    """Reporting the status of bridge and proxy, for debugging purposes"""
+    log(f"Is bridge running - {is_port_in_use(BRIDGE_PORT)}")
+    log(f"Is bridge proxy running - {is_port_in_use(BRIDGE_PROXY_PORT)}")
+
+
 def start(version: str, proxy: bool = False) -> None:
+    log("Starting")
     global proc
     global version_running
 
@@ -65,6 +83,9 @@ def start(version: str, proxy: bool = False) -> None:
     if proxy:
         bridge_proxy.start()
 
+    time.sleep(0.5)
+    check_bridge_and_proxy_status()
+
 
 def stop(proxy: bool = True) -> None:
     log("Stopping")
@@ -80,3 +101,6 @@ def stop(proxy: bool = True) -> None:
 
     if proxy:
         bridge_proxy.stop()
+
+    time.sleep(0.5)
+    check_bridge_and_proxy_status()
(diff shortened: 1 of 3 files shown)


grdddj:
Force-pushing again, as there was one commit which was not rebased on `incoming`

vdovhanych:
Good job, looks like it's working now. This [branch](https://github.com/trezor/trezor-suite/tree/vdovhanych/ci-tenv-artifacts-log) uses the latest test image and most of the [jobs](https://gitlab.com/satoshilabs/trezor/trezor-suite/-/pipelines/338562084) now pass

vdovhanych (approved):

grdddj:
Nice, thanks for running the CI and good job with the `debugging.log` artifact!

Seems it has 5 test scenarios inside (it received 5 requests with `"id": 0`), which would correspond to 5 integration testing stages - so it looks like it aggregated all the tests. It would be even more awesome if Suite tests could send a name/description of the test that is running and we would log it, so it can be quickly identified in the log file.

It is really helpful, as it prints the whole traceback of the error that failed the one test - lines 396-416 in the log file.

The test that failed is `./packages/integration-tests/projects/suite-web/tests/backup/t2-success.test.ts`, so I may try to run it locally to see if it is really problematic, or if it was a coincidence.

grdddj:
Tried the failing `backup/t2-success.test.ts` test locally three times and it was always fine, so there could be some accidental problem in CI

tsusanka:
The `backup/t2-success.test.ts` is problematic in CI for quite some time and we should fix it :-/. JFYI I know @mroz22 was doing some [modifications in Suite](https://github.com/trezor/trezor-suite/pull/3991) though that is not ready yet.

src/bridge.py
```diff
@@ -9,6 +11,9 @@
 import bridge_proxy
 import helpers
 
+BRIDGE_PORT = 21325
+BRIDGE_PROXY_PORT = 21326
```
tsusanka, line 15: `from .bridge_proxy import PORT as BRIDGE_PROXY_PORT`?

src/bridge.py
```diff
@@ -38,7 +43,21 @@ def get_status() -> dict:
...
 
+def is_port_in_use(port):
+    """Cheks if a certain port is listening on localhost"""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        return s.connect_ex(("0.0.0.0", port)) == 0
+
+
+def check_bridge_and_proxy_status() -> None:
+    """Reporting the status of bridge and proxy, for debugging purposes"""
+    time.sleep(0.5)
```
tsusanka, line 54: I think the sleep should be outside the function in `start` as it is not inherently needed from the point of the function. I mean we need it because we are starting/stopping but the function should be a standalone thing which doesn't need it. Does it make sense?

src/bridge_proxy.py
```diff
@@ -111,18 +31,18 @@ def start() -> None:
...
         return
 
-    SERVER = ThreadingServer((IP, PORT), Handler)
-    SERVER.daemon_threads = True
-    thread = threading.Thread(target=SERVER.serve_forever)
-    thread.daemon = True
-    thread.start()
+    file_path = os.path.join(os.path.dirname(__file__), "bridge_proxy_server.py")
+
+    command = f"python {file_path}"
```
tsusanka, line 36: This is quite ugly. Isn't there a better way?

tsusanka (approved):
I do not like how we are running `python` from `python` but I understand it. I think we can merge it, because TBH the bridge proxy is something we should tackle later on anyway. I am still not sure [this](https://github.com/trezor/trezor-user-env/pull/80/files#diff-66bfd2fb4185f66d394c44624273645e2bc41f80a46737ed5f2e54dd29790c1bR3-R8) is true and whether we should not solve it differently. But that's a topic for another time.

src/bridge.py
grdddj: Nice suggestion, as we actually already import `bridge_proxy`4 lines above this

grdddj:
Fixup commit 77744c8:
- importing bridge_proxy_port
- putting sleep outside of port checking function

tsusanka (approved):

grdddj:
Squashing the commits and rebasing to `incoming`. We can therefore merge if we want (and revisit the question of `what is the best way to run the proxy?` later)

PR #81, opened 2021-07-16
chore(docs): steps how to debug locally against Suite e2e tests

grdddj:
Documentation of how to run local Suite e2e tests against local trezor-user-env

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,3 +103,24 @@ This is suitable for smaller changes or things you can check via the HTML dashbo
 ### Let CI do it
 
 The simpler but less flexible way is to let the [CI](https://gitlab.com/satoshilabs/trezor/trezor-user-env/pipelines) build it. You can create a branch, commit your changes and then push them. The CI will build it for you and tag the appropriate docker image as `test`. You can then modify all scripts/commands and use `registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:test` instead of `registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env` which defaults to the `latest` tag which equals trezor-user-env's master. Suite's docker-compose files in the `docker` subdirectory are the place where you want to change this.
+
+## Local development against Suite end-to-end tests
+_NOTE: The steps below are all dependant on Docker containers. It could be even better to be able to avoid the Docker completely - just running trezor-user-env on localhost and connecting Suite tests to it - but hard to say how to do it, all my experiments failed. Please feel free to improve the solution._
+
+It may be helpful to run Suite e2e tests locally against the local trezor-user-env state, to avoid the need to involve the CI (Gitlab jobs) and the connected pain of merging changes into one branch just to test them.
+
+#### Suite preparation steps:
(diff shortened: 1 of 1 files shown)


tsusanka:
Thanks a lot for documenting this but there has to be a better (native) way to debug this instead of building docker images all the time.

What happens if you remove `trezor-user-env` from `docker/docker-compose.suite-test.yml` in Suite and start the servers locally? I mean the tests only depend on a services running on specific ports but they should not care whether that comes from Docker or not, no? :thinking:

grdddj:
It would be awesome if we could simplify it running `controller` just from localhost.

So far I tried to delete the container and its references from that `.yaml` and replace it by `host`. It had partial success, but some tests are unable to be run.

More experimenting/troubleshooting from my side needed

image

tsusanka:
Cc @mroz22 if by any chance you have some tips about this.

mroz22:
network mode host might interfere with some of your local services, typically bridge. but if you ensure that there is non running it might work I guess :thinking: 

Can't we declare trezor-user-env service like: 
```
trezor-user-env
  build: 
    context: .
    dokerfile: ../trezor-user-env/...
```
instead of using image from registry?

tsusanka:
IMHO we do not want to use docker image for trezor-user-env at all. We want to run `python main.py` directly so we do not need to rebuild all the time.

grdddj:
Fixup commit 3d1d90a:
- moving the NOTE about the current solution being sub-optimal to be more visible, and adding some extra info to it

README.md
```diff
@@ -91,3 +91,30 @@ This is suitable for smaller changes or things you can check via the HTML dashbo
...
+It may be helpful to run Suite e2e tests locally against the local trezor-user-env state, to avoid the need to involve the CI (Gitlab jobs) and the connected pain of merging changes into one branch just to test them.
+
+#### Suite preparation steps:
+- taken from [this](https://github.com/trezor/trezor-suite/blob/develop/docs/tests/e2e-web.md) and [this](https://github.com/trezor/trezor-suite/#development) Suite guides
+- `git clone git@github.com:trezor/trezor-suite.git`
+- `cd trezor-suite`
+- `sudo yarn`
+- `sudo yarn build:libs && yarn workspace @trezor/suite-data msg-system-sign-config`
+- `xhost +`
+- `docker/docker-suite-install.sh`
```
tsusanka, line 107: Maybe let's not copy and paste these commands and just link to the Suite's e2e-web.md file? In case it changes in future..

grdddj: These commands are working currently, so I would leave them here, to have everything in one place. We could add a note "in case of changes, the above-mentioned links show the up-to-date steps"

tsusanka: I really think it is better to just link the instructions. Because then for example we'll remove `msg-system-sign-config` in Suite and people will be super confused. Let's remove it.

grdddj:
Fixup commit 5c3f516:
- deleting the Suite preparation steps/commands, only leaving the links to up-to-date Suite documentation

tsusanka (approved):
Let's squash

grdddj:
> Let's squash

Should be ready

Issue #54, opened 2021-04-26
Pass all fields thru `emulator-apply-settings`


goodhoko:
The [apply settings](https://github.com/trezor/trezor-user-env/blob/619b5a10f72512f2cb65c34d36ca11352367e66d/controller/main.py#L106-L108) command currently forwards only a single field (`passphrase_always_on_device`) to the emulator. All other fields are silently ignored. This prohibits setting up an emulator with arbitrary features/settings for e2e tests.

_(TODO: @goodhoko link to usecase once a PR is posted.)_

Edit the websocket server so that it passes thru all fields to the emulator when `emulator-apply-settings` is invoked. Optimally make it fail loudly if the client sends a field that is unkown to the running emulator.

cc @tsusanka

goodhoko:
Thanks for addressing this @grdddj! I just noted that I forgot to link the usecase for this. It's the TODO's in this file: https://github.com/trezor/trezor-suite/blob/4a1207a7b9e72c08b224144206e65ba47479fc5b/packages/integration-tests/projects/suite-web/tests/suite/safety-checks-warning.test.ts

But I'd expect more e2e test could leverage this new capability. 😉

tsusanka:
@grdddj if you would like, feel free to contribute what @goodhoko is mentioning. If not we'll leave it for someone in the Suite team :).

Issue #18, opened 2020-10-07
Add multiple versions of trezorlib


tsusanka:
Because older firmwares might not work with the newest trezorlib. We can use virtualenv but we could also use Nix for that?

szymonlesisz:
especially `setup` and `decision` might not work

tsusanka:
Let's discuss this with matejcik.

tsusanka:
We will do https://github.com/trezor/trezor-user-env/issues/144 instead, we are not using the old firmwares anyway. Related: https://github.com/trezor/trezor-suite/issues/3728

PR #83, opened 2021-07-18
feat(controller): emulator-apply-settings takes all possible fields

grdddj:
Addressing issue #54:
- `emulator-apply-settings` command takes all possible fields, so the emulator can be more precisely set up for testing purposes
- when it receives an unsupported field, client will be notified, as we relay all the fields, and an unknown field would cause an Exception

NOTE(s):
- the `homescreen` field is a bytes object, so client should send it decoded as string (to be sent in JSON), and we will encode it into bytes
- the `experimental_features` field is not supported yet, as it seems like the current `trezorlib` version 0.12.2 is not yet compatible with it

### DIFF
--- a/src/controller.py
+++ b/src/controller.py
@@ -119,10 +119,14 @@ class ResponseGetter:
             emulator.wipe_device()
             return {"response": "Device wiped"}
         elif self.command == "emulator-apply-settings":
-            emulator.apply_settings(
-                self.request_dict["passphrase_always_on_device"],
-            )
-            return {"response": f"Applied setting on emulator {self.request_dict}"}
+            # Relaying all the relevant fields from the request, to make sure
+            #   the client is notified when it sends an unknown field
+            #   and the function will throw an Exception
+            settings_fields = deepcopy(self.request_dict)
+            settings_fields.pop("type", None)
+            settings_fields.pop("id", None)
+            emulator.apply_settings(**settings_fields)
+            return {"response": f"Applied settings on emulator {settings_fields}"}
         elif self.command == "emulator-reset-device":
             emulator.reset_device()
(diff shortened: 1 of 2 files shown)


src/controller.py
```diff
@@ -119,10 +119,14 @@ def run_command_and_get_its_response(self) -> dict:
             emulator.wipe_device()
             return {"response": "Device wiped"}
         elif self.command == "emulator-apply-settings":
-            emulator.apply_settings(
-                self.request_dict["passphrase_always_on_device"],
-            )
-            return {"response": f"Applied setting on emulator {self.request_dict}"}
+            # Relaying all the relevant fields from the request, to make sure
+            #   the client is notified when it sends an unknown field
+            #   and the function will throw an Exception
```
tsusanka, line 124: Not very important but we do not add the two spaces in the comments for next lines usually.

tsusanka (approved):
Looks good :+1:.

Regarding the homescreen: maybe it is time to start documenting the commands? But I have created a separate issue for that https://github.com/trezor/trezor-user-env/issues/84.

tsusanka:
I think I can merge, feel free to fix the comments styling any time later, but it is really nit-picking. Also we need to figure out the various `trezorctl` versions, but that is a topic for #18.

src/controller.py
grdddj: Sure, I will "fix" the comments, but I cannot promise I will not do it again in the future, as I do it automatically - I think it really quickens the text comprehension, one can immediately see that this "block" is one sentence (or logic block)

tsusanka: Feel free to keep it. I really do not have much of an opinion on that and you are right we should either enforce it in CI or forget about it :)).

Issue #86, opened 2021-07-20
Checking bridge status by HTTP GET does not work well when bridge is running


grdddj:
When bridge is running and we try to get its `/status/` endpoint, it usually takes (almost exactly) two seconds before it responds.

It creates issues with responsiveness, when `index.html` sends background status requests and "normal" commands are therefore blocked (the websocket has a lot of queued requests).

Proposed solution:
- change the way we are checking the bridge status (either check for port listening or for spawned process)

PR #87, opened 2021-07-20
fix(bridge): bridge status based on whether port listens or not

grdddj:
Addressing issue #86
- changing bridge status retrieving logic to just look whether the port is listening

### DIFF
--- a/src/bridge.py
+++ b/src/bridge.py
@@ -5,9 +5,6 @@ import socket
 import time
 from subprocess import Popen
 
-import requests
-from requests.exceptions import ConnectionError
-
 import bridge_proxy
 import helpers
 from bridge_proxy import PORT as BRIDGE_PROXY_PORT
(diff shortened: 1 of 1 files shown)


tsusanka (approved):

tsusanka:
That's very curious that it does take 2s for Bridge to respond doesn't it?

grdddj:
Yes, it is very unexpected, no idea why (also it could be just my local issue).

Also, when the `bridge` is running, there are some HTTP logs being printed into the console, even if it is being spawned in a separate process (and they are not connected to the background check we do from `index`).

Maybe my understanding of `bridge` is not complete - what is it "really" doing?

tsusanka:
Bridge is just a simple HTTP server on one side which receives requests and sends these via USB to the device. Repo: github.com/trezor/trezord-go/

tsusanka:
I mean this code removes code which is always nice :) so I am merging. If you would have some time in future feel free to investigate.

PR #89, opened 2021-07-21
Shamir

grdddj:
Second PR about Shamir backup walkthrough - testing that it works fine and adding comments to the functionality it does

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -52,7 +52,10 @@
   - **action**: simulates the Single backup process
 
 - **emulator-read-and-confirm-shamir-mnemonic**
-  - **action**: simulates the Shamir backup process
+  - **action**: simulates the Shamir backup process for chosen amount of shares and threshold
+  - **arguments**:
+    - **shares**: `int` (defaults to 1)
+    - **threshold**: `int` (defaults to 1)
 
 - **emulator-allow-unsafe-paths**
   - **action**: allow unsafe path on emulator
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -49,7 +49,10 @@
     - **value**: `str`
 
 - **emulator-read-and-confirm-mnemonic**
-  - **action**: simulates the backup process
+  - **action**: simulates the Single backup process
+
+- **emulator-read-and-confirm-shamir-mnemonic**
+  - **action**: simulates the Shamir backup process
 
 - **emulator-allow-unsafe-paths**
   - **action**: allow unsafe path on emulator
--- a/src/controller.py
+++ b/src/controller.py
@@ -107,6 +107,9 @@ class ResponseGetter:
         elif self.command == "emulator-read-and-confirm-mnemonic":
             emulator.read_and_confirm_mnemonic()
             return {"response": "Read and confirm mnemonic"}
+        elif self.command == "emulator-read-and-confirm-shamir-mnemonic":
+            emulator.read_and_confirm_mnemonic_shamir()
+            return {"response": "Read and confirm Shamir mnemonic"}
         elif self.command == "emulator-allow-unsafe-paths":
             emulator.allow_unsafe()
             return {"response": "Allowed unsafe path"}
--- a/src/controller.py
+++ b/src/controller.py
@@ -108,8 +108,14 @@ class ResponseGetter:
             emulator.read_and_confirm_mnemonic()
             return {"response": "Read and confirm mnemonic"}
         elif self.command == "emulator-read-and-confirm-shamir-mnemonic":
-            emulator.read_and_confirm_mnemonic_shamir()
-            return {"response": "Read and confirm Shamir mnemonic"}
+            shares = self.request_dict.get("shares") or 1
+            threshold = self.request_dict.get("threshold") or 1
+            emulator.read_and_confirm_shamir_mnemonic(
+                shares=shares, threshold=threshold
+            )
+            return {
+                "response": f"Read and confirm Shamir mnemonic for {shares} shares and threshold {threshold}."
+            }
         elif self.command == "emulator-allow-unsafe-paths":
             emulator.allow_unsafe()
             return {"response": "Allowed unsafe path"}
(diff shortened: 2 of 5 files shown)


docs/controller.md
```diff
@@ -49,7 +49,10 @@
     - **value**: `str`
 
 - **emulator-read-and-confirm-mnemonic**
-  - **action**: simulates the backup process
+  - **action**: simulates the classic backup process
```
tsusanka, line 52: Not that it matters much but we usually call it Single backup. Or technically BIP-39.

src/controller.py
```diff
@@ -107,6 +107,9 @@ def run_command_and_get_its_response(self) -> dict:
         elif self.command == "emulator-read-and-confirm-mnemonic":
             emulator.read_and_confirm_mnemonic()
             return {"response": "Read and confirm mnemonic"}
+        elif self.command == "emulator-read-and-confirm-mnemonic-shamir":
```
tsusanka, line 110: Maybe `emulator-read-and-confirm-shamir-mnemonic` better? Or `emulator-read-and-confirm-slip39-mnemonic`?

src/controller.py
```diff
@@ -107,6 +107,9 @@ def run_command_and_get_its_response(self) -> dict:
...
+            emulator.read_and_confirm_mnemonic_shamir()
+            return {"response": "Read and confirm mnemonic SHAMIR"}
```
tsusanka, line 112: `Read and confirm Shamir mnemonic`

tsusanka: Also we should use past tense here shouldn't we? So `Read and confirmed`. But feel free to do that in another PR.

src/emulator.py
```diff
@@ -228,37 +228,90 @@ def read_and_confirm_mnemonic() -> None:
...
+def read_and_confirm_mnemonic_shamir():
+    client = DebugLink(get_device().find_debug())
+    client.open()
     time.sleep(SLEEP)
 
-    client.swipe_up()
+    # Click Continue to begin Shamir setup process
+    client.press_yes()
+
+    # Clicking the minus button four times to set only one share (it starts at 5)
```
tsusanka, line 268: I do not know if we need to do it now (but maybe why not?) but the number of shares and threshold should be configurable. In another words we should have it as arguments to the `emulator-read-and-confirm-mnemonic-shamir` command?

src/controller.py
grdddj: I like the first one, `emulator-read-and-confirm-shamir-mnemonic`

src/controller.py
grdddj: Yes, good idea, will do it after this PR

src/emulator.py
grdddj: Agreed, it could be really helpful for Suite tests. I will try to play with it quickly

grdddj:
Fixup commit 32ede39:
- small changes in `shamir` wording

grdddj:
Commit 6e1b8ea:
- allowing for `shares` and `threshold` arguments into the Shamir backup walkthrough

src/emulator.py
```diff
@@ -252,34 +253,79 @@ def read_and_confirm_mnemonic() -> None:
...
     client.press_yes()
+    time.sleep(SLEEP)
+
     client.close()
 
 
-def read_and_confirm_mnemonic_shamir():
+def read_and_confirm_mnemonic_shamir(shares: int = 1, threshold: int = 1):
+    # TODO: decide what numbers will be allowed (there are some maximum numbers)
+    if shares < 1 or shares > 6:
```
tsusanka, line 269: As [per the spec](https://github.com/satoshilabs/slips/blob/master/slip-0039.md) there is at most 16 shares possible.

Btw the spec also allows something called SuperShamir. This is not however supported by Suite and won't be anytime soon. Could you still please share a note that "This function does not support Super Shamir"?

src/emulator.py
```diff
@@ -252,34 +253,79 @@ def read_and_confirm_mnemonic() -> None:
...
 
-    # Clicking the minus button four times to set only one share (it starts at 5)
-    minus_button_coords = (60, 70)
-    for _ in range(4):
-        client.click(minus_button_coords)
-        time.sleep(SLEEP)
-
-    # Click Continue to confirm one share
+    # Clicking the minus/plus button to set right number of shares (it starts at 5)
+    default_shares = 5
```
tsusanka, line 288: Let's have this as constant (therefore uppercase) as well.

src/emulator.py
```diff
@@ -252,34 +253,79 @@ def read_and_confirm_mnemonic() -> None:
...
+    # (threshold will be 1 and 2 respectively)
+    # Otherise assign it correctly
+    if shares not in [1, 2]:
+        # TODO: extend this for all supported numbers in the future
+        default_thresholds_for_shares_on_screen = {
+            3: 2,
+            4: 3,
+            5: 3,
+            6: 4,
+        }
```
tsusanka, line 317: We can't read this somehow via the debuglink as well right?

tsusanka (commented):
Just a few nit-picks. Btw the shamir function is getting quite complicated. It would be nice to test it. I have created a separate issue for that: https://github.com/trezor/trezor-user-env/issues/90.

src/emulator.py
grdddj: Good idea. Also, it seems to follow a pattern - `(x//2) + 1`, but we should verify it. I will try to take it from device, if there

grdddj: I could not find it written in the device directly, but [firmware repo](https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/management/reset_device/layout.py#L272) confirms the equation above, so I will use that

grdddj:
Fixup commit 3f5ec3d:
- creating some more CONSTANTS in the code
- using equation to get default threshold count on device instead of a hardcoded dictionary

src/emulator.py
tsusanka: Great :+1: :+1:

tsusanka (approved):
LGTM! Let's squash.

grdddj:
Doing second force-push just to include a forgotten "f" from a formatted string in controller response

tsusanka (approved):

Issue #82, opened 2021-07-17
Receive and log test scenario name/description from Suite e2e tests


grdddj:
It would be helpful for debugging purposes to send the test scenario name or small description from Suite tests to the websocket. Websocket would log this unique test identifier, so we would quickly know where to look in our logs.

Steps:
- create method in controller with a command like "log_test_details", receiving a string and logging it
- create Cypress task in Suite calling this websocket "endpoint"
- at the start of each Suite test, call this function to send unique test name/description (could also act as a nice extra documentation for that testing scenario)

tsusanka:
I am thinking we could even somehow enforce it. For example the client would have to send "hello" with these details and we would respond with [welcome](https://github.com/trezor/trezor-user-env/blob/f67ba1ca434dbcfaad95c0a4b1ba070a9dbb7e49/src/controller.py#L183). All other commands would raise an error until this is sent, what do you think? It seems better to enforce it otherwise it can get messy when debugging where one test sends it and the other does not and you do not know such feature in trezor-user-env.

grdddj:
If it could be enforced, even better. However, then the websocket would lose its stateless status, as it would need to remember which clients already sent this `hello` and which not.

Certainly something for discussion.

PR #93, opened 2021-07-22
Grdddj/log function for test details

grdddj:
Addressing suggestion issue #82:
- creating `log` function on `controller`, so that Suite tests (or any other client) can describe the test scenarios for debugging and auditing purposes

### DIFF
--- a/src/controller.py
+++ b/src/controller.py
@@ -65,6 +65,10 @@ class ResponseGetter:
         """Performs wanted action and returns details of what happened."""
         if self.command == "ping":
             return {"response": "pong"}
+        elif self.command == "log":
+            text = self.request_dict["text"]
+            log(text)
+            return {"response": "Text logged"}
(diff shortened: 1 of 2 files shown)


src/controller.py
```diff
@@ -65,6 +65,11 @@ def run_command_and_get_its_response(self) -> dict:
         """Performs wanted action and returns details of what happened."""
         if self.command == "ping":
             return {"response": "pong"}
+        elif self.command == "log":
+            text = self.request_dict["text"]
+            to_log = f"New test case: {text}"
```
tsusanka, line 70: I would omit the text and just log the `text`. Suite can add the `New test case: ` and it'll be able to use the `log` command for anything. Exactly as you write in the docs.

tsusanka (changes requested):

tsusanka (approved):

PR #97, opened 2021-07-27
feat(controller): command to click on specified emulator pixel coordination

grdddj:
Enabling clients to click anywhere on the emulator, based on inputted coordinations
- addressing the issue with [one Suite test](https://github.com/trezor/trezor-suite/blob/e1c710e0c6bbd96b0e7970c15e3650c092aa0c83/packages/integration-tests/projects/suite-web/tests/onboarding/t2-recovery-success.test.ts#L12), where it is somehow impossible to click the `Continue` button in newer emulator versions (bigger than 2.1.4) at the end of successful backup
- (yes, it is kind of hack, we should troubleshoot properly why `pressYes` is not working there, but this `emulator-click` can be used elsewhere as well)

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -48,6 +48,12 @@
   - **arguments**:
     - **value**: `str`
 
+- **emulator-click**
+  - **action**: click on a specified pixel coordination (x, y) on emulator
+  - **arguments**:
+    - **x**: `int`
+    - **y**: `int`
+
 - **emulator-read-and-confirm-mnemonic**
   - **action**: simulates the Single backup process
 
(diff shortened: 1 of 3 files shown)


tsusanka (approved):
We definitely want this, so :+1: and I'll merge but we should really investigate:
> impossible to click the Continue button in newer emulator versions (bigger than 2.1.4) at the end of successful backup

I think that should work. Maybe worth a firmware issue?

tsusanka:
I mean fee free to use the click function in Suite test but we should also create an issue to make sure this is under control.

grdddj:
I agree, it is something unexpected, will create issue in firmware. Not sure if there is a related issue in Suite, I made a quick search but could not locate it

Issue #99, opened 2021-07-30
Introduce regtest control


tsusanka:
In https://github.com/trezor/trezor-user-env/pull/98 @vdovhanych is adding a Bitcoin server in so-called [regtest](https://developer.bitcoin.org/examples/testing.html#regtest-mode) setup. This allows you to easily set up Bitcoin blockchain as needed (e.g. you can control mining).

We would like to add number of commands to tenv's _controller_ that would communicate with the regtest blockchain. As the bitcoind's [ports are exported](https://github.com/trezor/trezor-user-env/pull/98/files#diff-45bbb4f23e5ce9f69cefb061e34f0c64f59f4fcae4e13b4fce4408f45fb50d3bR26) we should be able to communicate with the bitcoind using the RPC interface.

Let's make a PoC of this feature. Let's start with a few commands:
- `bitcoin-generate-blocks(how many blocks)` - mine specific number of blocks
- `generatetoaddress(amount)` - make sure address contains `amount`. This might need more RPC commands than one.

grdddj:
Quick visualization:
- two containers inside one docker-compose structure
- controller receiving commands (same as now) and forwarding the messages to the second container through some port it can touch

![image](https://user-images.githubusercontent.com/42543243/127652681-2ddaedb2-11d9-4c51-ae8a-57b9a9140d6e.png)

What info we need:
- what commands will users want to send, and which parameters they will take
- how to translate these commands into RPC message that BTC regtest can understand

tsusanka:
> What info we need:

Let's start with the two commands mentioned above. I think just that will cover a lot of use cases.

PR #98, opened 2021-07-28
feat(ci): create bitcoid docker image and setup regtest mode

vdovhanych:
This pr is the first step to introducing new functionality to tenv. This adds bitcoin core running in regtest mode and also our blocbook service that is connected to there. 

This is still a work in progress. Just putting it here so anyone interested can track the changes.

PR #100, opened 2021-08-03
Grdddj/regtest controller

grdddj:
Addressing issue #99:
- connection to BTC regtest
- first two controller commands
- refactored controller commands into categories (emulator, bridge, regtest)

What next in regtest?:
- all depending on the usage by Suite tests
- we would probably need some read-only commands like `get_address_balance(address)` or `is_transaction_included(txid)` and functional commands like `announce_raw_transaction(tx_data)`


vdovhanych:
What happens when you have multiple wallets created? With multiple wallets, you always need to specify that wallet when you send some command to bitcoind.

grdddj:
> What happens when you have multiple wallets created? With multiple wallets, you always need to specify that wallet when you send some command to bitcoind.

I have not tried hat yet. Will it be our case that we will create multiple wallets and we will need to be able to operate all of them? I am looking into it

vdovhanych:
> > What happens when you have multiple wallets created? With multiple wallets, you always need to specify that wallet when you send some command to bitcoind.
> 
> I have not tried that yet. Will it be our case that we will create multiple wallets, and we will need to operate all of them? I am looking into it.

This I don't know. Can you have multiple addresses on there and only one wallet? We will want to check if the emulator completed the transactions.

grdddj:
> This I don't know. Can you have multiple addresses on there and only one wallet? We will want to check if the emulator completed the transactions.

One wallet can certainly have multiple (almost infinite amount) of addresses.

With testing emulator transactions I imagine the following process:
- emulator will itself generate one address
- we will send some BTC into that address on regtest
- we will send the UTXO details back into emulator
- emulator will create and sign a transaction spending that
- we will send the transaction details into regtest, announce this transaction and will mine a new block
- we will check that transaction is in the block

grdddj:
Can emulator work even with regtest/testnet addresses (they seem to be the same)? Regtest network does not support mainnet addresses

tsusanka:
> Can emulator work even with regtest/testnet addresses (they seem to be the same)?

Yes.

---

> With testing emulator transactions I imagine the following process:

- emulator will itself generate one address
- we will send some BTC into that address on regtest
- <s>we</s> Suite will send the UTXO details into emulator **as part of Suite e2e testing of normal transaction process**
- emulator will create and sign a transaction spending that
- <s>we</s>Suite will send the transaction details into <s>regtest</s> blockbook, **which will** announce this transaction to the regtest
- we will mine a new block
- we will check that transaction is <s>in the block</s> **displayed as mined in Suite**

docs/controller.md
```diff
@@ -108,5 +108,17 @@
 - **bridge-stop**
   - **action**: stop the bridge
 
+- **regtest-mine-blocks**
+  - **action**: mine specified amount of blocks, optionally by a specified address
```
tsusanka, line 112: What do you mean "by"? You mean where the coinbase reward goes to?

docs/controller.md
```diff
@@ -108,5 +108,17 @@
...
+  - **arguments**:
+    - **block_amount**: `int`
+    - **address**: `str` (optional) - defaults to the wallet's own address
+
+- **regtest-fill-address**
+  - **action**: send specified amount of BTC to specified address
```
tsusanka, line 118: Nit: I think we can drop the "specified" anywhere :).

src/controller.py
```diff
@@ -20,6 +21,26 @@ def log(text: str, color: str = LOG_COLOR) -> None:
...
+# TODO: think about some better solution how to make setup
+try:
+    REGTEST_RPC.createwallet("tenv_wallet")
+    log("Wallet created")
+except Exception as err:
+    if "Database already exists" not in str(err):
+        raise
+    log("Wallet already exists")
+FEES_PER_KVB = 0.00001  # 1 sat/vB
+REGTEST_RPC.settxfee(FEES_PER_KVB)
```
tsusanka, line 41: I think we should do this outside of tenv. @vdovhanych could we put this into the Dockerfile directly? Simply RUN some commands at the end?

src/controller.py
```diff
@@ -78,7 +99,47 @@ def run_command_and_get_its_response(self) -> dict:
...
+        elif self.command == "exit":
+            emulator.stop()
+            bridge.stop()
+            log("Exiting", "red")
+            exit(1)
+        elif self.command.startswith("bridge"):
+            return self.run_bridge_command()
+        # TODO: change "select-num-of-words" to "emulator-select-num-of-words"
+        # NOTE: needs to be changed in Suite tests as well
+        # (probably both should be supported for a while not to break old branches)
```
tsusanka, line 111: Let's create an issue instead please.

src/controller.py
```diff
@@ -151,26 +212,45 @@ def run_command_and_get_its_response(self) -> dict:
...
+
+    def run_regtest_command(self) -> dict:
+        if self.command == "regtest-mine-blocks":
+            block_amount = self.request_dict["block_amount"]
+            address = self.request_dict.get("address") or REGTEST_RPC.getnewaddress()
+            REGTEST_RPC.generatetoaddress(block_amount, address)
+            return {"response": f"Mined {block_amount} blocks by address {address}"}
+        elif self.command == "regtest-fill-address":
+            # NOTE: this is sending specified amount to address, if we wanted
+            #   to enforce the address has exactly X BTC, it would be harder
```
tsusanka, line 229: Maybe let's rename the command to `regtest-add-to-address`? Or `regtest-send-to-address`?

src/controller.py
```diff
@@ -151,26 +212,45 @@ def run_command_and_get_its_response(self) -> dict:
...
+            # (including fees and considering a big 1kvB transaction)
+            while True:
+                our_balance = REGTEST_RPC.getbalance()
+                if our_balance > btc_amount + FEES_PER_KVB:
+                    break
+                else:
+                    REGTEST_RPC.generatetoaddress(1, REGTEST_RPC.getnewaddress())
+                    log(
+                        f"Our balance {our_balance} is smaller than {btc_amount}. Mining a new block to increase it."
+                    )
```
tsusanka, line 243: This should be taken care of outside of tenv as well IMHO. @vdovhanych I think the container should be in a state where it is ready to be used directly. That means with a wallet created and with lot's of BTC.

src/bitcoin_rpc.py
```diff
@@ -0,0 +1,56 @@
+import decimal
```
tsusanka, line 1: Let's move this file to `src/bitcoin-regtest/rpc.py`.

tsusanka (changes requested):

docs/controller.md
grdddj: Yes, that this address will receive the newly mined BTC

src/controller.py
grdddj: Yes, I like the second one `regtest-send-to-address`

grdddj:
>   Suite will send the transaction details into blockbook, which will announce this transaction to the regtest

The `blockbook` - is it some layer between Suite and BTC network (does not matter if it is mainnet or regtest) that allows to check address balances, verify transactions, send them etc., so therefore we do not need any of this in our `tenv`?

src/controller.py
grdddj: Agreed, it would be nice, also from the point that this setup process in `controller` fails when the `regtest` container is not running. We would need some error handling, which could be avoided by doing that in the `regtest` itself

src/controller.py
vdovhanych: It should be possible to add it in the container from the startup. Will try to add it in the [pr.](https://github.com/trezor/trezor-user-env/pull/98)

grdddj:
Fixup commit 9335b28:
- addressing comments above

tsusanka:
> The blockbook - is it some layer between Suite and BTC network (does not matter if it is mainnet or regtest) that allows to check address balances, verify transactions, send them etc., so therefore we do not need any of this in our tenv?

Yes. And we have'll blockbook bundled in (https://github.com/trezor/trezor-user-env/pull/98).

src/controller.py
tsusanka: And let's remove the TODO.

tsusanka:
So this waits for the changes needed from @vdovhanych. Let's first merge #98, then @vdovhanych will make a new PR, we'll review that and then we can rebase and merge this one.

grdddj:
Rebased and force-pushed on top of ` bitcoin-regtest-server`, so that we have both the `controller` and `regtest image` code there at once, so I can try how they go together (and we see the exact git diff here)

grdddj:
Fixup commit 55273de:
- reflecting the changes in https://github.com/trezor/trezor-user-env/pull/103 in `controller` and synchronizing `controller` and `regtest image` parts

Issue #102, opened 2021-08-11
Create default wallet and fill it with coins in Bitcoin regtest


tsusanka:
As discussed [here](https://github.com/trezor/trezor-user-env/pull/100#discussion_r684813841) and [here](https://github.com/trezor/trezor-user-env/pull/100#discussion_r684814475).

PR #103, opened 2021-08-16
fix(ci): update regtest with default wallet and generate blocks

vdovhanych:
This PR addresses #102 

This seems to be the easiest solution to running the backend and filling it with bitcoins. As we need to wait a bit for the service to start, I had to add sleep, and in general, generating bitcoins will slow down the startup of the blockbook (approx 10-15 seconds) not sure if this could cause some problems with the tenv server @grdddj

### DIFF
--- a/docker/bitcoin-regtest/entrypoint.sh
+++ b/docker/bitcoin-regtest/entrypoint.sh
@@ -5,10 +5,25 @@ set -e
 cd /opt/coins/nodes/bitcoin_regtest
 
 echo "Starting bitcoin regtest backend service"
-/opt/coins/nodes/bitcoin_regtest/bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
+bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
+
+sleep 5
+
+# generate test wallet and blocks
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc createwallet "tenv-wallet"
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc -generate 150
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc settxfee 0.00001
 
 # start blockbook
 cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
-/opt/coins/blockbook/bitcoin_regtest/bin/blockbook -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db -sync -internal=:19021 -public=:19121 -certfile=/opt/coins/blockbook/bitcoin_regtest/cert/blockbook -explorer= -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs
+bin/blockbook \
+    -blockchaincfg=config/blockchaincfg.json \
+    -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
+    -sync \
+    -internal=:19021 \
+    -public=:19121 \
+    -certfile=cert/blockbook \
+    -explorer= \
+    -log_dir=logs


tsusanka (commented):
So tell me again why can't we use just `bin/bitcoin-cli` since we have `cd` into the folder?

Other then that I am assigning @grdddj and he'll check if this does the trick for him :).

vdovhanych:
> So tell me again why can't we use just `bin/bitcoin-cli` since we have `cd` into the folder?
> 
> 
> 
> Other then that I am assigning @grdddj and he'll check if this does the trick for him :).

Will shorten it, you're right it will work 😁

vdovhanych:
Force pushing shortened commands and changed amount of blocks generated so it's a bit faster to load.

tsusanka (approved):
LGTM but as mentioned let's wait for @grdddj

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -5,10 +5,16 @@ set -e
...
 
 echo "Starting bitcoin regtest backend service"
-/opt/coins/nodes/bitcoin_regtest/bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
+bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
+
+sleep 3
+
+# generate test wallet and blocks
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc createwallet "tenv-wallet"
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc -generate 1024 1000
```
grdddj, line 14: I would say that `-generate 150` is enough here, we do not need so many BTC (1024 blocks will yield 14700 BTC, 150 is the "halving" interval, and we get 2500 spendable BTC from it - more than enough).

The second argument is also optional, based on [docs](https://developer.bitcoin.org/reference/rpc/generate.html), we do not need to include it (unless there is a reason for it - @vdovhanych ?)

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -5,10 +5,16 @@ set -e
 cd /opt/coins/nodes/bitcoin_regtest
 
 echo "Starting bitcoin regtest backend service"
-/opt/coins/nodes/bitcoin_regtest/bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
+bin/bitcoind -datadir=/opt/coins/data/bitcoin_regtest/backend -conf=/opt/coins/nodes/bitcoin_regtest/bitcoin_regtest.conf &
```
grdddj, line 8: Here I am confirming that changing the config path to relative is really not possible, even though it is in the same dir as we `cd'd`

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -5,10 +5,16 @@ set -e
...
+# generate test wallet and blocks
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc createwallet "tenv-wallet"
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc -generate 1024 1000
 
 # start blockbook
 cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
-/opt/coins/blockbook/bitcoin_regtest/bin/blockbook -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db -sync -internal=:19021 -public=:19121 -certfile=/opt/coins/blockbook/bitcoin_regtest/cert/blockbook -explorer= -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs
+bin/blockbook -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db -sync -internal=:19021 -public=:19121 -certfile=/opt/coins/blockbook/bitcoin_regtest/cert/blockbook -explorer= -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs
```
grdddj, line 20: In this command, I was able to replace the absolute path with relative - deleting all the `/opt/coins/blockbook/bitcoin_regtest` prefixes of locations

grdddj:
>  generating bitcoins will slow down the startup of the blockbook (approx 10-15 seconds) not sure if this could cause some problems with the tenv server @grdddj

It is not so slow on my machine, I think it is alright. I did not try run both in containers with docker-compose yet, but I imagine the app must wait until both containers to initialize.

As for this PR, I am approving that. I made some smaller changes and additions to it in https://github.com/trezor/trezor-user-env/pull/100 (based on my comments here), so that the combination of `controller` and `regtest image` is there. We could migrate these `entrypoint` changes into this PR to make it separate from `controller`

grdddj (approved):

vdovhanych:
Yeah let's move it here or just merge this before your's and that should also be fine i guess no ?

grdddj:
> Yeah let's move it here or just merge this before your's and that should also be fine i guess no ?

Agreed, feel free to transfer the `entrypoint` changes here and merge, so my PR only touches `controller`

vdovhanych:
> > Yeah let's move it here or just merge this before your's and that should also be fine i guess no ?
> 
> Agreed, feel free to transfer the `entrypoint` changes here and merge, so my PR only touches `controller`

Ok

vdovhanych:
Force pushed with the changes you suggested. should be ready to go now.

grdddj:
> Force pushed with the changes you suggested. should be ready to go now.

Good, thanks for that. I agree we are ready to merge - my PR can follow (after review and green light)

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -5,10 +5,17 @@ set -e
...
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc createwallet "tenv-wallet"
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc settxfee 0.00001
+bin/bitcoin-cli -rpcport=18021 -rpcuser=rpc -rpcpassword=rpc -generate 150
 
 # start blockbook
 cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
-/opt/coins/blockbook/bitcoin_regtest/bin/blockbook -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db -sync -internal=:19021 -public=:19121 -certfile=/opt/coins/blockbook/bitcoin_regtest/cert/blockbook -explorer= -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs
+bin/blockbook -blockchaincfg=config/blockchaincfg.json -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db -sync -internal=:19021 -public=:19121 -certfile=cert/blockbook -explorer= -log_dir=logs
```
grdddj, line 21: Maybe last thing for discussion - what about splitting these long commands into multiple lines, like
```sh
bin/blockbook \
    -blockchaincfg=config/blockchaincfg.json \
    -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
    -sync \
    -internal=:19021 \
    -public=:19121 \
    -certfile=cert/blockbook \
    -explorer= \
    -log_dir=logs
```

Actually seeing this uncovered to me that `-explorer` is empty is it alright?

vdovhanych: Yes tahat should be Ok, all of the packages are run this way. Don't really think it's necessary to modify the commands any further I'm fine with how it is know but in all honesty whatever you guys like more I will be fine with.

vdovhanych:
Had to update the entrypoint again as it wasn't creating the wallet and blocks. Set it to sleep 5 and now it is consistent with creating that so now it should be ok.

docker/bitcoin-regtest/entrypoint.sh
tsusanka: Yeah that looks nicer. So @vdovhanych let's do this final minor change and ship it?

vdovhanych: Done, will test the docker image after it's built and if it's ok i will merge it then.

grdddj: I agree with merging

tsusanka: Yes, do it!

vdovhanych:
It works fine, merging it.

PR #104, opened 2021-08-23
chore(readme): update clone command so it's possible to clone even without github account

vdovhanych:
Updated the readme with git clone command to allow cloning for anyone even without any user account on Github. With ssh, you cant clone it that way.


tsusanka:
Merged via 8d6aa24 (hope you forgive me).

PR #105, opened 2021-08-26
Update README.md

Tondoshi:
Updated instructions for running on NixOS to be easier for non-tech users.


tsusanka:
I have added some info to our [Notion](https://www.notion.so/satoshilabs/trezor-user-env-7554614fa19a49f5ad47b16412016135) and I am closing here. We should not instruct users to `sudo` and Notion hopefully gives enough information on this.

PR #110, opened 2021-11-13
Fix/regtest config

szymonlesisz:
- use blockbook custom config. coin_shortcut: `REGTEST` instead of `rBTC`, compatible with[ trezor-common](https://github.com/trezor/trezor-common/blob/master/defs/bitcoin/bitcoin_regtest.json#L3)

- disable unnecessary https in blockbook service, now its served at `http://127.0.0.1:19121/`

### DIFF
--- a/docker/bitcoin-regtest/blockchaincfg.json
+++ b/docker/bitcoin-regtest/blockchaincfg.json
@@ -1,0 +1,20 @@
+{
+    "coin_name": "Regtest",
+    "coin_shortcut": "REGTEST",
+    "coin_label": "Regtest",
+    "rpc_url": "http://127.0.0.1:18021",
+    "rpc_user": "rpc",
+    "rpc_pass": "rpc",
+    "rpc_timeout": 25,
+    "parse": true,
+    "message_queue_binding": "tcp://127.0.0.1:48321",
+    "subversion": "",
+    "address_format": "",
+    "xpub_magic": 70617039,
+    "xpub_magic_segwit_p2sh": 71979618,
+    "xpub_magic_segwit_native": 73342198,
+    "slip44": 1,
+    "mempool_workers": 8,
+    "mempool_sub_workers": 2,
+    "block_addresses_to_keep": 300
+}
--- a/docker/bitcoin-regtest/entrypoint.sh
+++ b/docker/bitcoin-regtest/entrypoint.sh
@@ -19,11 +19,11 @@ cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
 bin/blockbook \
-    -blockchaincfg=config/blockchaincfg.json \
+    -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json \
     -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
     -sync \
     -internal=:19021 \
     -public=:19121 \
-    -certfile=cert/blockbook \
     -explorer= \
-    -log_dir=logs
+    -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs \
+    -dbcache=0
--- a/docker/bitcoin-regtest/Dockerfile
+++ b/docker/bitcoin-regtest/Dockerfile
@@ -16,6 +16,8 @@ RUN wget https://github.com/vdovhanych/blockbook/releases/download/v0.3.5/backen
   && rm blockbook-bitcoin-regtest_0.3.5_amd64.deb \
   && rm backend-bitcoin-regtest_0.21.1-satoshilabs-1_amd64.deb
 
+# replace default blockbook config with custom config with changed coin_shortcut to "REGTEST", to be compatibile with trezor-common shorcut.
+COPY ./docker/bitcoin-regtest/blockchaincfg.json /opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json
 COPY ./docker/bitcoin-regtest/entrypoint.sh ./entrypoint.sh
 RUN chmod +x ./entrypoint.sh
 


docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -19,11 +19,11 @@ cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
 bin/blockbook \
-    -blockchaincfg=config/blockchaincfg.json \
+    -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json \
```
tsusanka, line 22: @vdovhanych do we need the full path actually?

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -19,11 +19,11 @@ cd /opt/coins/blockbook/bitcoin_regtest
...
     -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
     -sync \
     -internal=:19021 \
     -public=:19121 \
-    -certfile=cert/blockbook \
     -explorer= \
-    -log_dir=logs
+    -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs \
+    -dbcache=1073741824
```
tsusanka, line 29: Is this needed?

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -19,11 +19,11 @@ cd /opt/coins/blockbook/bitcoin_regtest
...
-    -blockchaincfg=config/blockchaincfg.json \
+    -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json \
     -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
     -sync \
     -internal=:19021 \
     -public=:19121 \
-    -certfile=cert/blockbook \
     -explorer= \
-    -log_dir=logs
+    -log_dir=/opt/coins/blockbook/bitcoin_regtest/logs \
```
tsusanka, line 28: @vdovhanych would be nice to have this log somehow easily accessible. But let's worry about that later..

docker/bitcoin-regtest/blockchaincfg.json
```diff
@@ -0,0 +1,21 @@
...
+    "rpc_timeout": 25,
+    "parse": true,
+    "message_queue_binding": "tcp://127.0.0.1:48321",
+    "subversion": "",
+    "address_format": "",
+    "xpub_magic": 70617039,
+    "xpub_magic_segwit_p2sh": 71979618,
+    "xpub_magic_segwit_native": 73342198,
+    "slip44": 1,
+
```
tsusanka, line 17: Unnecessary newline. But don't worry about it if it'll be the only change, I'll just fix it later.

docker/bitcoin-regtest/entrypoint.sh
```diff
@@ -19,11 +19,11 @@ cd /opt/coins/blockbook/bitcoin_regtest
 
 echo "Starting blockbook service"
 bin/blockbook \
-    -blockchaincfg=config/blockchaincfg.json \
+    -blockchaincfg=/opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json \
     -datadir=/opt/coins/data/bitcoin_regtest/blockbook/db \
     -sync \
     -internal=:19021 \
     -public=:19121 \
-    -certfile=cert/blockbook \
```
tsusanka, line 27: Aaah so that's all that was needed to disable HTTPS? Aaah 🤦

cc @karliatto maybe you know already but after we merge this, trezor-user-env will contain this change and blockbook will be accessible via plain HTTP.

tsusanka (approved):
LGTM, let's wait for @vdovhanych just to be sure.

docker/bitcoin-regtest/entrypoint.sh
martinboehm: Absolutely not. This is ridiculous cache size for the regtest. I would use `-dbcache=0`

szymonlesisz: sorry, i've copied id from somhere else :) 
531042e

docker/bitcoin-regtest/blockchaincfg.json
szymonlesisz: 50eb2c6

docker/bitcoin-regtest/entrypoint.sh
szymonlesisz: i've put it here just to b sure that valid config file will be used, see Dockerfile
```
COPY ./docker/bitcoin-regtest/blockchaincfg.json /opt/coins/blockbook/bitcoin_regtest/config/blockchaincfg.json
```
it was not obvious to me where `config/blockchaincfg.json` is pointing

szymonlesisz: maybe it belongs to previous commit, idk... or i can cange it back

docker/bitcoin-regtest/entrypoint.sh
vdovhanych: We could extract it as artifact in the gitlab pipeline in the same way we do with the tenv logs. Could be done later.

docker/bitcoin-regtest/entrypoint.sh
vdovhanych: i think yes, as it's launching it from the bin and uses config that is somewhere else. I think that is should or hast to be full path.

vdovhanych (approved):
This looks nice, LGTM.

docker/bitcoin-regtest/entrypoint.sh
tsusanka: Will you open an issue pls?

tsusanka (approved):

PR #114, opened 2021-11-19
arm and m1 apple silicon support

vdovhanych:
This introduces support for M1 or any other arm device that supports docker. This pr is still a draft since there need to be some more changes. 

TODO: 

- [x] Add arm device in our CI for building the emulator 
- [x] Add arm device in our CI for building arm Docker image. Alternatively, set it up with `docker buildx`
- [x] Change firmware script to fetch latest arm emulator build from CI

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -34,6 +34,39 @@ build and push test image:
     - docker push $CONTAINER_NAME:$CI_COMMIT_SHA
     - docker push $CONTAINER_NAME:test
 
+# trezor-user-env arm docker images build
+.build_arm: &build_arm
+  image: $CI_DEPENDENCY_PROXY_GROUP_IMAGE_PREFIX/docker
+  variables:
+    CONTAINER_NAME_ARM: "$CI_REGISTRY/satoshilabs/trezor/trezor-user-env/trezor-user-env-arm"
+  services:
+    - docker:dind
+  before_script:
+    - docker login $CI_REGISTRY -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD
+
+build and push arm image:
+  stage: build
+  <<: *build_arm
+  only:
+    - master
+    - vdovhanych/m1-support
+  script:
+    - docker build --tag $CONTAINER_NAME_ARM:latest  -f docker/arm/Dockerfile .
+    - docker push $CONTAINER_NAME_ARM:latest
+
+build and push arm test image:
+  stage: build
+  <<: *build
+  except:
+    - schedules
+    - master
+    - /^run\//
+  script:
+    # note that this way "test" tag gets overwritten everytime
+    - docker build --tag $CONTAINER_NAME_ARM:$CI_COMMIT_SHA --tag $CONTAINER_NAME_ARM:test -f docker/arm/Dockerfile .
+    - docker push $CONTAINER_NAME_ARM:$CI_COMMIT_SHA
+    - docker push $CONTAINER_NAME_ARM:test
+
 # regtest docker images build
 .build_regtest: &build_regtest
   image: $CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX/docker
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -36,20 +36,20 @@ build and push test image:
 
 # trezor-user-env arm docker images build
 .build_arm: &build_arm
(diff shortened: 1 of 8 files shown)


tsusanka:
OMG IT WORKS 🎉 🎉 🎉 

image

tsusanka:
I have dared to modify the docs a bit :).

vdovhanych:
Let's close this 😢. It should work out of the box now with the latest docker app for m1, so no need for this 🙈

tsusanka:
Upon further discussion we came to the conclusion this will be more stable and makes sense to finish. We will not worry about old firmwares: we will start supporting ARM from the latest release and if one needs older FWs they simply run it via rosette.

vdovhanych:
This should be finished and ready to be merged with a tiny exception. We need to change the download link for master emulators once the changes are merged to master there. We decided to finish this and use it for now. We should be using more native solutions on ARM macs, which could be solved by switching to nix native environment. Let's discuss this next year @tsusanka.

tsusanka:
trezor/trezor-firmware#2025 merged!

vdovhanych:
> [trezor/trezor-firmware#2025](https://github.com/trezor/trezor-firmware/pull/2025) merged!

@tsusanka This should be good to go then.

ci/build.yml
```diff
@@ -34,6 +34,38 @@ build and push test image:
...
+  stage: build
+  <<: *build
+  when: manual
+  except:
+    - schedules
+    - master
+    - /^run\//
+  script:
+    - docker build  --tag $CONTAINER_NAME_ARM:CI_COMMIT_BRANCH-test -f docker/arm/Dockerfile .
+    - docker push $CONTAINER_NAME_ARM:CI_COMMIT_BRANCH-test
```
tsusanka, line 67: Let's remove this, we are reworking this in https://github.com/trezor/trezor-user-env/pull/127 anyway.

tsusanka: You meant `<<: *build_arm` anyway :).

Let's remove the anchor altogether then.

docker/arm/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM vdovhanych/nixos
+
+# "fake" dbus address to prevent errors
+# https://github.com/SeleniumHQ/docker-selenium/issues/87
+#ENV DBUS_SESSION_BUS_ADDRESS=/dev/null
```
tsusanka, line 5: Let's remove this.

src/binaries/trezord-go/bin/download.sh
```diff
@@ -5,3 +5,4 @@ set -e
 
 # Bridge needs an older glibc so we are pinning to nixos-21.05 (stable) as of 2021-07-02
 nix-shell -p autoPatchelfHook -I "nixpkgs=https://github.com/NixOS/nixpkgs/archive/e9148dc1c30e02aae80cc52f68ceb37b772066f3.tar.gz" --run "autoPatchelf trezord-*"
+nix-shell -p autoPatchelfHook  --run "patchelf --set-interpreter /nix/store/0y64vbl8hc0xqq5ag6wwj5r3pr0l6z2s-glibc-2.33-56/lib/ld-linux-aarch64.so.1 ./src/binaries/trezord-go/bin/*-arm64"
```
tsusanka, line 8: Hardcoding the path is definitely wrong. We can't autopatch it? @mmilata might help here.

tsusanka (commented):
I have added some comments. This will need more love in future but as a PoC it is great :).

ci/build.yml
vdovhanych: removed it in c0220b4

docker/arm/Dockerfile
vdovhanych: removed in c0220b4

src/binaries/trezord-go/bin/download.sh
vdovhanych: I think if we change it to `nix-shell -p autoPatchelfHook  --run "patchelf --set-interpreter /lib/ld-linux-aarch64.so.1 ./src/binaries/trezord-go/bin/*-arm64"` it should work and use the current store path

vdovhanych: Ok so we can't, it's not working without setting it like that.

vdovhanych:
Closing this as we decided that multiarch build of docker image would be better.

PR #117, opened 2021-11-23
chore(nix): updating channel before we patch emulator files

### LINKED ISSUE #116: 2-master won't start (2.4.3)
trying to start 2-master (2.4.3) started to produce this error
```
/trezor-user-env/src/binaries/firmware/bin/trezor-emu-core-v2-master: /nix/store/sbbifs2ykc05inws26203h0xwcadnf0l-glibc-2.32-46/lib/libc.so.6: version `GLIBC_2.33' not found (required by 
```

cc @tsusanka

vdovhanych:
This should fix launching the latest master version emulator. The problem was with different GCC lib versions and simple nix-channel update fixed that as we already were using unstable channel in the base docker image of tenv. 
fixes #116

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,6 +22,9 @@ WORKDIR ./trezor-user-env
 RUN ls /
 RUN ls /trezor-user-env
 
+# updates the nix-channel to have latest changes
+RUN nix-channel --update
+
 RUN nix-shell --run "./src/binaries/firmware/bin/download.sh"
 RUN nix-shell --run "./src/binaries/trezord-go/bin/download.sh"
 


tsusanka:
@mmilata why this change matter if we have our shell.nix pinned? 🤔 

https://github.com/trezor/trezor-user-env/blob/d45d9940d7f657d3117de229fe7e7d133e9ef02c/shell.nix#L6

vdovhanych:
@tsusanka as for me I think that it's because we launch `nix-shell -p ` for patching and that may use the default set channel for this (just guessing) 

We should still update the nix packages in the `shell.nix` version to the same as we have on firmware but that brought some error in the code check. Ci run here - https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/1810921504

tsusanka:
Aaah that's it! Ehh, that seems kind of dull :). I think more correct solution is to have everything in shell.nix. I will merge this now, so we have the master fixed and will improve it tonight.

> We should still update the nix packages in the shell.nix version to the same as we have on firmware but that brought some error in the code check. Ci run here - https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/1810921504

This is due to new mypy. I'll check it out.

PR #119, opened 2021-11-24
Bump nix packages to the same commit as in Firmware

### LINKED ISSUE #116: 2-master won't start (2.4.3)

tsusanka:
I have remembered why it is done this way: you can't use the autoPatchHook directly in the shell.nix for some reason. So I have pinned the nix-shell command to the same commit but is still kind of suboptimal that we need to pin three times.

@mmilata is there a simple other way to solve this? Just in case there is something missing otherwise it is not important I would say.

Closes #116


src/binaries/firmware/bin/download.sh
```diff
@@ -40,7 +40,7 @@ cd "$BIN_DIR"
 
 # mark as executable and patch for Nix
 chmod u+x trezor-emu-*
-nix-shell -p autoPatchelfHook SDL2 SDL2_image --run "autoPatchelf trezor-emu-*"
+nix-shell -p autoPatchelfHook SDL2 SDL2_image -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/7fad01d9d5a3f82081c00fb57918d64145dc904c.tar.gz --run "autoPatchelf trezor-emu-*"
```
mmilata, line 43: Perhaps you can set `NIX_PATH` variable in shell.nix? Should work the same as `-I`.

docker/Dockerfile
```diff
@@ -22,9 +22,6 @@ WORKDIR ./trezor-user-env
 RUN ls /
 RUN ls /trezor-user-env
 
-# updates the nix-channel to have latest changes
-RUN nix-channel --update
```
vdovhanych, line 26: i suppose we could keep this in case we run in to other version discrepancies when running other stuff for example. But i'll leave it up to you.

vdovhanych (approved):

src/binaries/firmware/bin/download.sh
tsusanka: Yup, this seems to work! Added in https://github.com/trezor/trezor-user-env/pull/119/commits/5e3894ca30015c21fc81ab5fdfe1a8276a77307e.

docker/Dockerfile
tsusanka: I guess you are right, it doesn't hurt. The bad thing is we won't know now whether the pinning works or not, but let's find out later :). I put it back.

tsusanka:
So please recheck this should ready from my side :).

mmilata:
Looks good! I refactored it a bit to not repeat the commit hash: https://github.com/trezor/trezor-user-env/commit/2db48b04daecc250c63c2cf1ed21539a2c65dd96, feel free to cherrypick.

Also it's quite un-obvious but all top-level attributes passed to `mkDerivation` are exported as env variables.

tsusanka:
Very nice! Cherry-picked and force-pushed, thx. @vdovhanych last round of review on your side and let's merge?

vdovhanych (approved):
looks good :)

tsusanka:
Ok, I am merging as this should not break anything™️. Thx!

tsusanka:
It seems I broke it? 🤔 https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/1851419221#L3735

tsusanka:
For some reason the `nix-shell --run "./src/binaries/trezord-go/bin/download.sh"` script does not patch trezord-go correctly. However, the very same equivalent for firmware works just fine. Weird..

tsusanka:
Ok so ironically Bridge needs to stay at the old version.. Seems like https://github.com/trezor/trezor-user-env/pull/121 fixes it. But let's keep this PR as well as it seems as a correct solution for now.

Issue #49, opened 2021-01-28
Use custom firmware


szymonlesisz:
Currently we are working only with prefetched FW, it would be nice to have a possibility to work with custom build or custom `trezor-firmware/branch` or custom `trezor-firmware-fork/branch`

Usecase: testing connect against not merged branches/pull requests in trezor-firmware repo

matejcik:
user story: I have a local checkout of the firmware repo with some things built. I want to run Suite e2e tests via trezor-user-env, against the locally built thing.

Options:
- i launch the emulator locally, tell the docker image to use it
- i launch the emulator and bridge locally, tell the docker image to use my bridge instead of launching its own
	* advantage: this way I can run Suite tests against a hardware device
- i submit the emulator binary to docker image, tell it to use it
	* drawback: this only works for frozen builds
- i tell the docker image to use my local checked-out directory
    * advantage: I can use non-frozen builds (which need to see the Python sources in the filesystem)
    * disadvantage: tenv probably doesn't know how to launch emulator in this configuration? (you need to be at the right working directory)

grdddj:
We have agreed to try the **option 3**, where the binary file is already available (non-frozen build will need to be compiled, but it is a quick process).

Probably main question right now is: **do we want to run the whole thing locally or using CI?** (we can support even both)
- it can be theoretically rephrased as **do we want to run only specific tests with new firmware, or all of them?** (only specific -> locally is better, all of them -> CI is better)
  - _Running the Suite e2e test locally can be a pain. I do not know about a good way how to run all the tests locally at once, without clicking each test individually in Cypress window. And even if it was possible, it would probably take a VERY long time, as in CI every stage runs in parralel, not sure if it can be done locally_

Steps needed to be done in every case:
- having the binary emulator file (in case of TT, it is around 11 MB currently, so can be quite easily uploaded almost everywhere)
- `tenv` having access to it
- `Suite` e2e tests specifying this version

**CI route**
- binary emulator can be uploaded somewhere (into Gitlab)
- CI job could take the binary URL as an argument, store it in `Suite` e2e test code, which will be forwarding that to `tenv` with a request to spawn this emulator (sending that URL instead of `2-master)`
- `tenv` will recognize this is not a normal version and will try to download it from that URL and then run it (it will need to download it only once, then `tenv` will know it already has it)

**Local route**
[...]

PR #120, opened 2021-12-03
feat(emulator): allowing to download and run emulator from URL

grdddj:
Connected with https://github.com/trezor/trezor-firmware/issues/1932 and https://github.com/trezor/trezor-user-env/issues/49

Ability to specify emulator URL which will be downloaded and run by `tenv`.

Tested using emulators from `Gitlab CI` - 
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1839236800/artifacts/raw/core/build/unix/trezor-emu-core
https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1839236825/artifacts/raw/legacy/firmware/trezor.elf

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -27,8 +27,15 @@
   - **action**: start the specified version of emulator (and if one already runs, kills it)
   - **arguments**:
     - **version**: `str` (1.9.4, 2.4.0., etc.) - default is the latest TT
-    - **wipe**: `bool` ... whether to delete the emulator profile before starting it - default is False
+    - **wipe**: `bool` whether to delete the emulator profile before starting it - default is False
     - **output_to_logfile**: `bool` whether the debug output should go to a logfile - default is True - otherwise it goes to stdout
+
+- **emulator-start-from-url**
+  - **action**: downloads emulator from specified URL and runs it
+  - **arguments**:
+    - **url**: `str` from where to download the emulator
+    - **model**: `str` which emulator it is - either "1" for T1 or "2" for T2
+    - **wipe**: `bool` whether to delete the emulator profile before starting it - default is False
 
 - **emulator-stop**
   - **action**: stop the emulator
--- a/src/dashboard/index.html
+++ b/src/dashboard/index.html
@@ -38,24 +38,40 @@
 
     <section>
         <h3>Emulator</h3>
-        <div class="explain-note">Use left and right arrow keys to emulate the Trezor One buttons. Mouse clicks for Trezor T.</div><br>
-        <div id="emu-status-line"><b>Status: <span id="emu-status">unknown</span></b></div>
+        <div class="explain-note">
+            Use left and right arrow keys to emulate the Trezor One buttons. Mouse clicks for Trezor T.
+        </div>
+
+        <br>
+
+        <div id="emu-status-line">
+            <b>Status: <span id="emu-status">unknown</span></b>
+        </div>
         <div>
             <h4>Trezor One</h4>
             <select id="t1-select"></select>
             <button onclick="emulatorStart('t1-select');">Start</button>
-            <br />
         </div>
         <div>
             <h4>Trezor T</h4>
             <select id="t2-select"></select>
             <button onclick="emulatorStart('t2-select');">Start</button>
         </div>
-        <input id="wipeDevice" type="checkbox">
-        <label for="wipeDevice">Start with wiped/empty device</label><br>
-        <input id="emulatorUseLogfile" type="checkbox">
-        <label for="emulatorUseLogfile">Logs into logfile</label><br>
-        <hr />
+        <div>
+            <input id="wipeDevice" type="checkbox">
+            <label for="wipeDevice">Start with wiped/empty device</label>
+            <input id="emulatorUseLogfile" type="checkbox">
+            <label for="emulatorUseLogfile">Logs into logfile</label><br>
+        </div>
+        <div>
+            <input id="emu-url" type="text" placeholder="Full emulator URL, link to Gitlab job or just job ID - as specified in select. Do not forget to specify model (T1/T2)" style="width:50%;"/>
+            <select id="emu-url-format-select"></select>
+            <select id="emu-url-model-select"></select>
+            <button onclick="emulatorStartFromUrl();">Start from URL</button>
+        </div>
+
+        <hr>
+
         <h3>Emulator commands</h3>
         <div class="explain-note">These commands are auto-confirmed using the emulator's debug link.</div><br>
         <div>
--- a/src/binaries.py
+++ b/src/binaries.py
@@ -46,8 +46,8 @@ def explore_firmwares(args: Any) -> None:
 
 
 def sort_firmwares(version: str) -> tuple:
-    # Having master version as the first one in the list
-    if "master" in version:
+    # Having master and url versions as the first one in the list
+    if "master" in version or "url" in version:
         return 99, 99, 99
     return tuple(int(n) for n in version.split("."))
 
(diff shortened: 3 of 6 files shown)


docs/controller.md
```diff
@@ -29,6 +29,13 @@
     - **version**: `str` (1.9.4, 2.4.0., etc.) - default is the latest TT
     - **wipe**: `bool` ... whether to delete the emulator profile before starting it - default is False
 
+- **emulator-start-from-url**
+  - **action**: downloads emulator from specified URL and runs it
+  - **arguments**:
+    - **url**: `str` from where to download the emulator
+    - **one_or_two**: `str` which emulator it is - either "1" for T1 or "2" for T2
```
tsusanka, line 36: Let's use "model" instead?

docs/controller.md
```diff
@@ -29,6 +29,13 @@
...
+    - **wipe**: `bool` ... whether to delete the emulator profile before starting it - default is False
```
tsusanka, line 37: Huge detail but let's remove the dots (...) ?

tsusanka (commented):
Awesome 🎉.

There is one thing I do not know: do we want to support generic URLs or are we fine with Gitlab links? In other words do we want to support downloads from anywhere?

Because if we limit it for Gitlab, we can simplify some code and also the UI. We would fill the artifacts names automatically. We could also request the job ID (1839236800) only.

Do we see use-cases where you would want to use some **URL outside of Gitlab**? cc @matejcik @mroz22 

We could of course do both..

---

Also this is a great start and if we want to support branches easily we can do that by crafting a special URL. This URL `https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/artifacts/master/download?job=core%20unix%20frozen%20debug%20build` always leads to the latest core build on master. Replacing the master with any branch will do the same trick. But let's finish this PR first and worry about that then.

docs/controller.md
grdddj: I agree, let's rename it. Also, we can consider what is the best values it could take - it can be strings `"1"` and `"2"`, integers `1` and `2`, strings `"T1"` and `"T2"/"TT"` etc. Current situation is compatible with the first character in version -  `1.2.3` or `2.3.4`

matejcik:
To be specific: although perhaps convenient in some situations, I never need to upload a new emulator while tenv is running. A well documented `docker cp` would work just fine, together with version version number "locally supplied" parallel to `master`.
 -- that, or the ability to specify local path at runtime (but the URL thing would work with a file path too, maybe?)

docs/controller.md
grdddj: Renamed in 430eec3

grdddj:
> Is there a usecase that actually wants tenv to download a specified URL?

This would allow the firmware `CI` (where the artifacts get created) and any tests using `tenv` (Suite, Connect) being independent (running on different pipelines, environments, etc.). Also for the manual testing it could be useful so that `QA` or anybody else can test the behavior of any emulator being built at `Gitlab` or anywhere else.

matejcik:
> This would allow the firmware `CI` (where the artifacts get created) and any tests using `tenv` (Suite, Connect) being independent (running on different pipelines, environments, etc.).

supplying a local file would work just as well -- the CI pipeline instances are not talking to the _same_ docker instance running tenv, are they?

> Also for the manual testing it could be useful so that QA or anybody else 

Yes, this seems nice. In that case we can do what @tsusanka suggests and accept a link to the job instead of to the artifact.
I'm not sure how much it helps though. It's a UX nicety but other than that the code is essentially the same?

grdddj:
> To be specific: although perhaps convenient in some situations, I never need to upload a new emulator while tenv is running. A well documented `docker cp` would work just fine, together with version "locally supplied" parallel to `master`. -- that, or the ability to specify local path at runtime (but the URL thing would work with a file path too, maybe?)

I did some prototype with `docker cp` some time ago in https://github.com/trezor/trezor-suite/pull/4107 - after starting the `tenv` container we can copy any emulators into it and use them as `2-master` - alternatively giving them some special name like `2-custom`. We can theoretically do it in runtime as well, however probably only by pushing the emulator inside again by `docker cp`, I do not think `docker` container could get files from the local machine freely (but I would be glad to be mistaken)

matejcik:
I see, I must have missed that. It looks like it would work fine for our CI usage. Specific name like `custom` would be a better solution than `X-master` (probably not even including version, because tenv can't know and verify that it's the right version), but that's a cosmetic thing.

grdddj:
> the CI pipeline instances are not talking to the _same_ docker instance running tenv, are they?

Each pipeline using `tenv` is getting the image from `registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:latest`, so I think they _do not_ use the same instance, just the same base image.

grdddj:
> It looks like it would work fine for our CI usage.

So we could wait after all the jobs in `build CI` stage are finished, take the `core/build/unix/trezor-emu-core` file, transfer it into the `tenv` as `2-custom` and tell the `Connect` tests to use `2-custom` tests?

`Connect`'s main testing script, https://github.com/trezor/connect/blob/develop/tests/run.sh, already has the `-f` option to specify the emulator version, so we can supply that `2-custom` there.

I will be trying to implement this pipeline as our `CI` stage in https://github.com/trezor/trezor-firmware/pull/1972

We should probably test is against T1 emu as well, but it has lower priority (as it is not changing so much)

matejcik:
not all, just the one that produces the file we need (so presumably "core unix debug")
otherwise, yes, exactly.

tsusanka:
Sooo if I sum it up: we will do FW tests using simple `cp` as we'll require the build job using `needs` so the build will be present inside the container. This is nice and simple so 👍. @grdddj tackles this in https://github.com/trezor/trezor-firmware/pull/1972 so let's continue the discussion there.

----

Now regarding this PR: could we still make use of this URL feature? Exactly as you say I can imagine some use cases for QA or Suite devs where you need to test some particular branch (like a Taproot one recently). It seems like a nice feature but it is true that we should not implement something until we properly understand the need. @mroz22 what do you think of this? Would you find it useful if you could set Gitlab URL in tenv and simply launch the emulator from such pipeline?

tsusanka:
Let's change the URL to https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/1839236825/ and fill in the details

src/dashboard/index.html
```diff
@@ -38,24 +38,42 @@ <h3>Bridge</h3>
...
-        <hr />
+        <div>
+            <input id="wipeDevice" type="checkbox">
+            <label for="wipeDevice">Start with wiped/empty device</label>
+            <input id="emulatorUseLogfile" type="checkbox">
+            <label for="emulatorUseLogfile">Logs into logfile</label><br>
+        </div>
+        <div>
+            <!-- TODO: could be prefilling the t1-or-t2-select based on some keywords in URL
+            (like "trezor-emu-core" fills T2, "trezor.elf" fills T1)  -->
```
tsusanka, line 68: Let's remove the comment, I think the select box is enough.

tsusanka (approved):
One tiny thing, feel free to force-push then :). Did not test, but let's see how it works :).

grdddj:
> One tiny thing, feel free to force-push then :). Did not test, but let's see how it works :).

Force-pushed after squashing everything

src/dashboard/index.html
grdddj: Removed in the force-pushed commit (force-pushed twice, second already rebased on current master, even though there was no conflict)

tsusanka:
I am getting a conflict now. Probably due to #126.

grdddj:
> I am getting a conflict now. Probably due to #126.

Correct, there was a need of rebase. Force-pushed now

PR #123, opened 2021-12-08
redirecting debug logs from emulator and bridge into a logfile

grdddj:
Connected with issues in https://github.com/trezor/trezor-firmware/pull/1972 with enormous log in `Gitlab` disabling us from seeing the final result of `Connect` tests and discussion with @matejcik 

Sending all the output from `emulator` and `bridge` into `logs/emulator_bridge.log`.

We could make this conditional on some parameter sent into `bridge.start()` and `emulator.start()` functions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ emulator.img
 .mypy_cache
 docker/version.txt
 logs/debugging.log
+logs/emulator_bridge.log
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -28,6 +28,7 @@
   - **arguments**:
     - **version**: `str` (1.9.4, 2.4.0., etc.) - default is the latest TT
     - **wipe**: `bool` ... whether to delete the emulator profile before starting it - default is False
+    - **output_to_logfile**: `bool` whether the debug output should go to a logfile - default is True - otherwise it goes to stdout
 
 - **emulator-stop**
   - **action**: stop the emulator
@@ -104,6 +105,7 @@
   - **action**: start the specified version of bridge (only if it is not already running)
   - **arguments**:
     - **version**: `str` (2.0.27, 2.0.31, etc.) - defaults to the latest available one
+    - **output_to_logfile**: `bool` whether the debug output should go to a logfile - default is True - otherwise it goes to stdout
 
 - **bridge-stop**
   - **action**: stop the bridge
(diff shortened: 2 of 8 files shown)


matejcik:
why not have separate logfiles for bridge and for emulator?

a better way to redirect is specifying the `stdout` parameter for `Popen` call, see https://github.com/trezor/trezor-firmware/blob/master/python/src/trezorlib/_internal/emulator.py#L135
pass `None` to not redirect.

fwiw tenv should ideally be using the above (`trezorlib._internal.emulator`) as a module instead of implementing the same thing locally. but that's not for this PR

grdddj:
> why not have separate logfiles for bridge and for emulator?

Also a possibility, but I would say for troubleshooting purposes the combination of those is better than two separate files - all the traffic can be seen at one place and in chronological order

grdddj:
In 63ff017 deciding to use above-mentioned option 2 - getting rid of that ugly `popen_kwargs`, also because `mypy` was complaining about the call to `Popen` - https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/1863683252

tsusanka:
I am also inclined having this in one file. I think, it should not be the default option however, because when you use as a developer you want to see it in stdout. Don't you think?

grdddj:
> I am also inclined having this in one file. I think, it should not be the default option however, because when you use as a developer you want to see it in stdout. Don't you think?

Making it (the redirect to a logfile) default makes sure the `Gitlab` will not show a lot of these (not very important) logs in `CI` pipelines - `Suite`, `Connect` and now also `Firmware` - without any needed code change in these repositories. If it was _not_ default, we would need to explicitly specify `output_to_logfile: True`.

I haven't used the `emulator` and `bridge` logs very frequently - and if needed, developers can just disable that `output_to_logfile` or `tail -f` the logfile to see realtime traffic (actually when searching for something, it is much easier in a file that in stdout)

vdovhanych:
I would rather have the logs in one file than two separate ones. Some errors could be connected, and comparing two files would be unnecessary "pain".  Agree that these logs are not needed for the pipeline output, and if there is a need for that, we could do it per job basis.

grdddj:
> Because I really like that you see what is happening when playing around with it right away without some need of fiddling around with it:

If you are using `tenv`'s `Dashboard` to spawn the emulators, then we could just specify `logfile: False` in `index.js`, and all emulators and bridges spawned from there would go to `stdout`.

This seems better and less invasive than having to explicitly specify `logifle: True` in all the `CI` pipelines

grdddj:
Force-pushed after rebase on `master`. Also, checkboxes created in `Dashboard` for `emu` and `bridge`, so it can be chosen whether logs go to `logfile` or not

tsusanka (approved):

PR #124, opened 2021-12-08
chore(emulator): better handling of allow_unsafe

grdddj:
Coming from discussion with @matejcik in https://github.com/trezor/trezor-firmware/pull/1972#issuecomment-987813473

When calling `allow_unsafe`, catching the possible error of not being able to apply it. The benefit is that any caller of `tenv` (like `connect` in our case) will not have to care about not calling it for lower versions not supporting safety checks.

Additionally we could check that the error message contains either `Unsupported safety-checks setting` (case for `T1`) or `No setting provided` (case for older `T2`). (Not to catch for example `Device is not initialized` when trying to allow checks on fresh device)

@matejcik from when does `T1` support safety checks (and does it really)? I am getting the errors even on `1.10.2`, which is relatively new

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -431,7 +431,7 @@ def allow_unsafe() -> None:
     client.open()
     time.sleep(SLEEP)
 
-    # Each model has its own supported safety checks level
+    # T1 does not support PromptAlways
     if client.features.major_version == 1:
         safety_checks = messages.SafetyCheckLevel.PromptTemporarily
     else:
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -5,9 +5,9 @@ from pathlib import Path
 from subprocess import PIPE, Popen
 from typing import Optional
 
-from trezorlib import debuglink, device
+from trezorlib import debuglink, device, messages
 from trezorlib.debuglink import DebugLink, TrezorClientDebugLink
-from trezorlib.device import reset, wipe
+from trezorlib.exceptions import TrezorFailure
 from trezorlib.transport.bridge import BridgeTransport
 from trezorlib.transport.udp import UdpTransport
 
@@ -168,7 +168,7 @@ def wipe_device() -> None:
     client = TrezorClientDebugLink(get_device())
     client.open()
     time.sleep(SLEEP)
-    wipe(client)
+    device.wipe(client)
     client.close()
 
 
@@ -176,7 +176,7 @@ def reset_device() -> None:
     client = TrezorClientDebugLink(get_device())
     client.open()
     time.sleep(SLEEP)
-    reset(client, skip_backup=True, pin_protection=False)
+    device.reset(client, skip_backup=True, pin_protection=False)
     client.close()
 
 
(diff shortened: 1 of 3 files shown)


matejcik:
> from when does `T1` support safety checks (and does it really)

since 1.10.1 but only `PromptTemporarily` value, not `PromptAlways`

grdddj:
> since 1.10.1 but only `PromptTemporarily` value, not `PromptAlways`

Alright, so we will need to send `safety_checks=2` in case of `T1`?

tsusanka (approved):
Great, good idea 👍. LGTM we only need to solve the last @matejcik's comment. I guess we'll have to use Temporarily for T1

grdddj:
> I guess we'll have to use Temporarily for T1

In 0428631 I made it like that (without using a `SafetyCheckLevel` enum, which seems to be not there in the currently released version of `trezorlib`) and it is fine.

I am confirming that `T1` supports safety checks from `1.10.1` and `T2` from `2.3.2` (as `Connect` was saying)

tsusanka:
@grdddj I think @matejcik just released new trezorlib version. Do you think you could update it in another PR and then rebase this on top that after?

@matejcik there are no breaking changes in trezorlib right? It seems we should be fine? https://github.com/trezor/trezor-firmware/blob/master/python/CHANGELOG.md#incompatible-changes Old emulators need to continue working.

tsusanka:
> @grdddj I think @matejcik just released new trezorlib version. Do you think you could update it in another PR and then rebase this on top that after?

I have just realized we can't do that by simply updating shell.nix as 0.13.0 is not there yet. We can either wait or switch to poetry as we use in firmware?

grdddj:
> I have just realized we can't do that by simply updating shell.nix as 0.13.0 is not there yet. We can either wait, add it or switch to poetry. I think adding poetry could be beneficial in future as well?

I like the idea of using `poetry` here as well, the benefit of having new python packages immediately available is nice. I will do it in a separate `PR` then

grdddj:
Force-pushed after rebasing on fresh `master` with `Poetry`, so we could use the `messages.SafetyCheckLevel.XXX` enum from new `trezorlib`

grdddj:
Force-pushed with `Allow unsafe (safety checks)` button text in `Dashboard` - together with small design change - putting command buttons on two rows, as there is too many

src/emulator.py
```diff
@@ -428,10 +428,27 @@ def apply_settings(
...
+        device.apply_settings(client, safety_checks=safety_checks)
+    except TrezorFailure as err:
+        # Catching only specific error message, otherwise reraising the exception
+        if "No setting provided" in str(err):
+            log(
+                f"Could not allow unsafe. Device does not support safety checks. Err: {err}",
+                "red",
+            )
+        else:
+            raise
```
tsusanka, line 453: This doesn't have to be `raise err`? I honestly do not know 🤔

src/emulator.py
```diff
@@ -428,10 +428,27 @@ def apply_settings(
...
 def allow_unsafe() -> None:
     client = TrezorClientDebugLink(get_device())
-    # ignore for Legacy firmware, there is no such setting
-    if client.features.major_version == 1:
-        return
     client.open()
     time.sleep(SLEEP)
-    device.apply_settings(client, safety_checks=1)  # TODO
+
+    # Each model has its own supported safety checks level
```
tsusanka, line 434: To be precise T1 does not support PromptAlways, TT supports all three of them (though not all of them are in Suite).

tsusanka (commented):
Just tiny questions and we are good to go :)

src/emulator.py
matejcik: no :)

src/emulator.py
grdddj: Is it alright just to replace this comment by `T1 does not support PromptAlways`?

grdddj: Done in 27200b1

tsusanka (approved):

tsusanka:
I have squashed and merged directly :). Thanks!

Issue #125, opened 2021-12-08
Take the emulator operation logic from firmware


grdddj:
As suggested in https://github.com/trezor/trezor-user-env/pull/123#issuecomment-988911436, we could use already existing, thorough and well-tested logic of handling emulators (starting, stopping etc.) instead of creating it on our own in `tenv`

PR #126, opened 2021-12-09
chore(emulator): emulator start/stop taken from trezorlib._internal.emulator

grdddj:
Implements https://github.com/trezor/trezor-user-env/issues/125
- reusing already existing logic of `start`ing and `stop`ing emulator from `trezorlib`

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -1,11 +1,12 @@
 import os
-import signal
+import sys
 import time
 from pathlib import Path
 from subprocess import PIPE, Popen
 from typing import Optional
 
 from trezorlib import debuglink, device, messages
+from trezorlib._internal.emulator import CoreEmulator, LegacyEmulator
 from trezorlib.debuglink import DebugLink, TrezorClientDebugLink
 from trezorlib.exceptions import TrezorFailure
 from trezorlib.transport.bridge import BridgeTransport
@@ -15,10 +16,11 @@ import bridge
 import helpers
 
 # TODO: consider creating a class from this module to avoid these globals
-proc = None
 version_running = None
+EMULATOR = None
 
 ROOT_DIR = Path(__file__).parent.parent.resolve()
+FIRMWARE_DIR = ROOT_DIR / "src/binaries/firmware/bin"
 
 # When communicating with device via bridge/debuglink, this sleep is required
 #   otherwise there may appear weird race conditions in communications.
@@ -86,10 +88,10 @@ def get_status() -> dict:
 
 
 def start(version: str, wipe: bool, output_to_logfile: bool = True) -> None:
-    global proc
     global version_running
+    global EMULATOR
 
-    if proc is not None:
+    if EMULATOR is not None:
         log(
             f"Before starting a new emulator - version {version}, "
             f"killing the already running one - version {version_running}",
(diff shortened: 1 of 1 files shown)


src/emulator.py
```diff
@@ -87,58 +88,49 @@ def get_status() -> dict:
...
-        command = f"{path}/trezor-emu-core-v{version} -O0 -X heapsize=20M -m main"
+        EMULATOR = CoreEmulator(
+            FIRMWARE_DIR / f"trezor-emu-core-v{version}", profile_dir=FIRMWARE_DIR
+        )
+    elif version[0] == "1":
+        os.environ["TREZOR_OLED_SCALE"] = str(TREZOR_ONE_OLED_SCALE)
+        EMULATOR = LegacyEmulator(
+            FIRMWARE_DIR / f"trezor-emu-legacy-v{version}",
+            profile_dir=FIRMWARE_DIR,
+            extra_args=["-O0"],
```
matejcik, line 111: AFAIK the `legacy` emulator does not need nor use this argument. Not sure why it was there originally, perhaps so that you could run `core` and `legacy` with the same command?

matejcik (approved):

src/emulator.py
grdddj: Thanks for pointing this, I have deleted it. It really could be that previously there was same command for `core` and `legacy`

grdddj:
Force-pushed after rebase on `master` with `output_to_logfile` functionality - need to specify `sys.stdout` as an argument to `Emulator`, otherwise the logs are not shown anywhere

grdddj:
> This needs rebase and is good to go then :).

Rebased and force-pushed

src/emulator.py
```diff
@@ -86,70 +88,57 @@ def get_status() -> dict:
...
             "red",
         )
         stop()
 
-    path = ROOT_DIR / "src/binaries/firmware/bin"
+    # Conditionally redirecting the output to a logfile instead of terminal/stdout
+    if output_to_logfile:
+        logfile = open(helpers.EMU_BRIDGE_LOG, "a")
+    else:
+        logfile = sys.stdout
```
tsusanka, line 106: Isn't this some bad copy&paste? Or should it really be here? I am just a bit confused it was not part of the previous PR

grdddj: It needs to be here now, as we are spawning emulator a different way. Previously it was with explicit `subprocess.Popen`, and when we do not supply `stdout` and `stderr` kwargs into it, it will automatically choose existing stdout and stderr.

However, current `LegacyEmulator` and `CoreEmulator` is _always_ calling the `stdout` and `stderr` arguments in `subprocess.Popen`, so we need to supply _either_ the logfile, or "current" stdout, otherwise the logfile will be set internally as `self.logfile = self.profile_dir / "trezor.log"`

Details at 
https://github.com/trezor/trezor-firmware/blob/30a77a76bc41296a32269694a666fadfc79d89ec/python/src/trezorlib/_internal/emulator.py#L69-L72
and
https://github.com/trezor/trezor-firmware/blob/30a77a76bc41296a32269694a666fadfc79d89ec/python/src/trezorlib/_internal/emulator.py#L132-L138

Issue #122, opened 2021-12-05
Test trezor-user-env


tsusanka:
We should add tests to this CI that will:
- Launch Bridge
- Launch master emulators
- Launch some specific emulator version

Because otherwise if we mess up something, Suite and other job start to fail and it can get very confusing.

grdddj:
I would try to do it in this sprint, as I am already doing some smaller `tenv` improvements for the `Connect` integration into firmware `CI`. It also means it would be another usage of `tenv`, so possibly another thing to break when `tenv` malfunctions.

Hard to say what is the best way of testing.
- we could just call the functions directly  - `emulator.start()`, `bridge.start()`, but that does not catch any issue in `controller`
- we would probably want to test the whole request/response pipeline, which would mean spinning up the `controller` and creating some mock client sending commands into it and expecting non-error responses
- we do have `bridges` in repository,  so we can spin them immediately, however, we would need to be downloading the `emulators` every time (certainly not all of them, just the `master` ones and some other "random" ones to try)
- test in `docker` container or not? It would probably take a lot of time and resources to always build it - but it could catch `docker` issues
  - we could have this "advanced" testing only in `master`, which is anyway already building `tenv:latest` images per new commit
  - it could be a `postbuild` stage, being dependent on the built image and verifying it runs fine

Just a momentary idea:
- we could have some `health-check` function that would try all possible things (launching bridge, emu, etc.) and all the users of `tenv` would/could call it before using it - and when it would fail, we would not continue with the tests and we would know the current `tenv:latest` is broken

grdddj:
Seems it can be connected with https://github.com/trezor/trezor-user-env/issues/90

grdddj:
This is theoretically also finished - via https://github.com/trezor/trezor-user-env/pull/127

**Overview of the testing**

Quick way how to test a recent production image (but theoretically any instance running locally):
- running `tenv` - `./run.sh`
- spawning tests - `make test_with_running_controller`

There is also `make test`, which spawns `tenv` from the current python codebase, runs the tests, and kills it

tsusanka:
Great 🎉. So the only thing we are missing is to kill Gitlab and move this to GH Actions correct?

vdovhanych:
Yes that should be it

tsusanka:
Moving to https://github.com/trezor/trezor-user-env/issues/163 then 🎉.

Issue #90, opened 2021-07-21
Add tests


tsusanka:
As some of our functions are getting quite large (e.g. the one for Shamir) we should consider adding tests for the trezor-user-env itself. The idea is:

- Create a dir `tests`.
- Setup `pytest` testing.
- Add it to CI (new `test` stage).
- Write first test, we could test the Shamir first, why not, that means: call the websocket, start the emulator, start intialization using `trezorctl reset-device` and go through the Shamir Backup.

It might seem a bit redundant to test both here and in Suite, but having tests here for the <s>websocket</s> controller will ensure everything is working on the trezor-user-env side in case we run into some failures in Suite CI.

PR #127, opened 2021-12-10
chore(tests): basic tests for controller commands

grdddj:
Connected with https://github.com/trezor/trezor-user-env/issues/90 and https://github.com/trezor/trezor-user-env/issues/122

So far just some basic tests of `controller`, trying to call most of its commands and checking if we receive a response with `success: True` (or `success: False` in case of purposely failing commands)

The `controller` currently has to run on the background (be started manually)

Next steps/ideas:
- automatically spawn the `controller` before the tests and kill it after them
- include `pytest`
- include it into `CI`
- some advanced functions are omitted now, as they are relying on specific emulator state - `"emulator-read-and-confirm-mnemonic"` and `emulator-read-and-confirm-shamir-mnemonic`
- could include the `regtest` - but that would require spinning up the `regtest` container

### DIFF
--- a/ci/test.yml
+++ b/ci/test.yml
@@ -1,0 +1,20 @@
+image: $CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX/nixos/nix:2.3.12
+
+test:
+  stage: test
+  needs: []
+  variables:
+    SDL_VIDEODRIVER: "dummy"
+  script:
+    - nix-shell --run "./src/binaries/firmware/bin/download.sh && ./src/binaries/trezord-go/bin/download.sh"
+    - nix-shell --run "poetry install"
+    - nix-shell --run "poetry run make test"
+  after_script:
+    - cp /builds/satoshilabs/trezor/trezor-user-env/logs/debugging.log trezor-user-env-debugging.log
+    - cp /builds/satoshilabs/trezor/trezor-user-env/logs/emulator_bridge.log tenv-emulator-bridge-debugging.log
+  artifacts:
+    paths:
+      - trezor-user-env-debugging.log
+      - tenv-emulator-bridge-debugging.log
+    expire_in: 1 week
+    when: always
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: "Controller tests"
+
+on: [pull_request]
+
+
+jobs:
+  controller_test:
+    runs-on: ubuntu-latest
+    env:
+      SDL_VIDEODRIVER: "dummy"
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - name: Setup Nix
+        uses: cachix/install-nix-action@v13
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+
+      - name: Download emulator and bridge binaries
+        run: nix-shell --run "./src/binaries/firmware/bin/download.sh && ./src/binaries/trezord-go/bin/download.sh"
+      - name: "Install poetry"
+        run: nix-shell --run "poetry install"
+      - name: Run the tests
+        run: nix-shell --run "poetry run make test"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,8 +4,10 @@ variables:
 
 stages:
   - prebuild
+  - test
   - build
 
 include:
   - ci/prebuild.yml
+  - ci/test.yml
   - ci/build.yml
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,9 @@ code_check:
 	@mypy .
 	@echo [FLAKE8]
 	@flake8 .
+
+test:
+	pytest -s
+
+test_with_running_controller:
+	pytest -s --controller-already-runs
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -110,6 +110,9 @@
 
 - **emulator-reset-device**
   - **action**: reset the device
+  - **arguments**:
+    - **use_shamir**: `bool` (optional) - whether to use Shamir backup (SLIP39) - default is False (BIP39 will be used)
+
 
 - **emulator-get-screenshot**
   - **action**: get current screen encoded as base64
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,4 +1,26 @@
 [[package]]
+name = "atomicwrites"
+version = "1.4.0"
+description = "Atomic file writes."
+category = "main"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+
+[[package]]
+name = "attrs"
+version = "21.4.0"
+description = "Classes Without Boilerplate"
+category = "main"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+
+[package.extras]
+dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit", "cloudpickle"]
+docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
+tests = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "cloudpickle"]
+tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "cloudpickle"]
+
+[[package]]
 name = "black"
 version = "21.12b0"
 description = "The uncompromising code formatter."
@@ -110,6 +132,14 @@ optional = false
 python-versions = ">=3.5"
 
 [[package]]
+name = "iniconfig"
+version = "1.1.1"
+description = "iniconfig: brain-dead simple config-ini parsing"
+category = "main"
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "isort"
 version = "5.10.1"
 description = "A Python utility / library to sort Python imports."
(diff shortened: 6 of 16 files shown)


grdddj:
After some troubleshooting on `CI` side (it is crucial to set `SDL_VIDEODRIVER: "dummy"` to run emulator on headless server), we have the first green `test` stage in https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/pipelines/427059570

What is still a mystery, is that `bridge` seems not to be spawned, probably we will need to call that `download.sh` script for it as well, even though it is already in repo.

Under the current conditions, this setup does not even need `pytest`, we are using simple `asserts` without it. However, when/if we will have more tests and more complex testcases, it would be probably nice to have it from the point of `parametrizing` the tests and not having to put all tests in one function.

ci/test.yml
```diff
@@ -0,0 +1,17 @@
+image: $CI_DEPENDENCY_PROXY_GROUP_IMAGE_PREFIX/nixos/nix:latest
+
+test:
+  stage: test
```
vdovhanych, line 4: Connected to the first comment I would run this In the stage after we build the image and test the image that was built.

grdddj: (Together with that first comment of yours):

That is right that there might be slightly different behavior in production image - so in `master` branch, we should really test the whole `tenv:latest` image after being built.

However, for normal branches, I would leave only these quicker tests in that `basic nixos docker` image, as that does not need any building from our side (and is therefore much quicker and less resource-demanding)

grdddj:
Current situation includes calling all non-regtest commands from `controller` and expecting a `success` response (however, not all possible arguments are being tried).

It now also tests the correct behavior of `read_and_confirm_mnemonic` and `read_and_confirm_shamir_mnemonic`, which both need a special setup for itself, and therefore are in their own functions. (We are using `trezorctl device backup` to initiate the backup process, which is not very nice however)

Hard to say how/whether to continue with other test cases, we could be testing all possible combinations of commands and expecting some errors to be thrown (like `device already initialized`), but this is probably not a `tenv`-thing.

@tsusanka @vdovhanych Do you have any ideas, what other functionality could be tested?

grdddj:
Force-pushed after rebase on `master`

ci/test.yml
vdovhanych: Agree. Let's do it this way then.

vdovhanych:
I think that for the basic functionality of tenv, this is sufficient for now, and we can maybe see if some situations will require additional testing.

ci/test.yml
```diff
@@ -0,0 +1,18 @@
+image: $CI_DEPENDENCY_PROXY_GROUP_IMAGE_PREFIX/nixos/nix:latest
```
vdovhanych, line 1:
```suggestion
image: $CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX/nixos/nix:latest
```
This is related to #132

vdovhanych (approved):
One small change but otherwise its looking good.

grdddj:
> I think that for the basic functionality of tenv, this is sufficient for now, and we can maybe see if some situations will require additional testing.

Cool. The last commit trying to run all the tests in freshly-build image before pushing it is not tested yet, due to current CI issues. Apart from that, it is all tested both locally and by CI.

We discussed with @tsusanka that we may add support for `pytest`in the future. However I do not see the need now, it would complicate things - in the current condition it enough having a python script with asserts. This way the script can also accept CLI arguments (for example `--controller-already-runs` as we have now). The `parametrization` by `pytest` would not bring much benefits, as the sequence of steps in `commands_success` needs to be called in the exact order (we need to spawn emulator at the beginning and set it up to try all the emulator commands).

ci/build.yml
```diff
@@ -17,6 +17,9 @@ build and push image:
     - /^run\//
   script:
     - docker build --tag $CONTAINER_NAME:$CI_COMMIT_SHA --tag $CONTAINER_NAME:latest -f docker/Dockerfile .
+    - docker-compose -f docker/compose.yml up trezor-user-env-unix &
+    - nix-shell --run "poetry install"
+    - nix-shell --run "poetry run make test_with_running_controller"
```
tsusanka, line 22: I do not like this. I understand the point of it, but we should not test in the build stage, it is confusing when it fails. I would remove it.

ci/test.yml
```diff
@@ -0,0 +1,44 @@
...
+    - cp /builds/satoshilabs/trezor/trezor-user-env/logs/debugging.log trezor-user-env-debugging.log
+    - cp /builds/satoshilabs/trezor/trezor-user-env/logs/emulator_bridge.log tenv-emulator-bridge-debugging.log
+  artifacts:
+    paths:
+      - trezor-user-env-debugging.log
+      - tenv-emulator-bridge-debugging.log
+    expire_in: 1 week
+    when: always
+
+test image:
```
tsusanka, line 22: I think the right way to test the image is by running the tests directly inside. That means use the `image` directive and set the docker image to the previously built one.

docker/compose.yml
```diff
@@ -7,6 +7,13 @@ services:
     environment:
       - DISPLAY=:0
 
+  trezor-user-env-unix-test:
+    container_name: trezor-user-env-test.unix
+    image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:test
+    network_mode: "host"
+    environment:
+      - DISPLAY=:0
```
tsusanka, line 15: And we would not need this at all then

tsusanka (commented):
Two main comments:

## Pytest
Please let's add pytest now already. I know it seems redundant but these things are better to solve right away, because then someone else comes it seems redundant to them as well and then we have lots of non-pytest tests and that would be a pity :). 


## CI workflow

I think the correct workflow should be like this:

For non-master branches:
1. prebuild phase
2. "native" tests (that means in a nixOS image)
3. docker image building as a manual job only

For master:
1. prebuild phase
2. build phase - builds docker image but tags it with a commit id
3. test phase - test inside the previously built docker image. Use the commit id as a tag and run the tests directly inside using the `image` directive.
4. deploy phase - rename the commit id tag to latest. Delete the commit id tag.

We could add the "native" tests here as well, but as they run inside the docker image AND they run natively as part of a pull requests, I think it is redundant.

ci/build.yml
grdddj: Yes, I agree, it is not very good. Done by suggested process of using the `image` directive using the freshly-built `tenv` image.

One (small) disadvantage of running tests _inside_ the image - we are _not_ using the image exactly as it is used in production. Let's say `Suite` is downloading the image and running it by `docker-compose` and communicates with exposed ports. Here we are not connecting to the running container at all, so we are not testing the `docker` exposes ports and we do not test the `CMD` instruction that is executed _only_ when the container runs (we need to replicate it by hand in our testing code).

Not sure if that makes sense and if it is even a "problem" :)

grdddj:
> Two main comments:
> ## Pytest
> 

Agreed, in the end, it was beneficial to transfer it. You are right, `Pytest` can also accept arguments, it was used in `firmware` already.

I was afraid of the parametrization, but it turns out it still maintains the order of tests, so I could reuse the same vectors. We _could_ maybe spawn `emulator` before each test and kill it afterwards, but current state is fine (running it at the beginning of the vectors and stopping it at the end) - we test the `emulator` start/stop in a different testcase.

> ## CI workflow
> 

Currently we have these 5 stages
```
  - prebuild
  - test
  - build
  - image_test
  - deploy
```
, where only the first two will be run for `non-master` branches, build (including push of `test` tag) is optional.

Last two are only run for `master` (and it will not run the test one). Currently for developing purposes I am not pushing `latest`, but `test` - and running it also for my `grdddj/tenv_tests` branch. After we agree on this, we need to change it to be production-ready.

Example successful pipeline with correctly pushed `test` image is at https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/pipelines/442029388

The names of the stages are open, we could maybe rename `test` to `native_test` to separate it by name from `image_test`

tsusanka:
I forgot about this one a little bit, I'll make a final review? Or are there other blockers?

ci/build.yml
```diff
@@ -8,17 +8,20 @@
...
+# Builds image and pushes it only with "$CI_COMMIT_SHA" tag
+# Push with "latest" tag (in deploy stage) will only be done after
+# the image is tested (in image_test stage)
+build image:
   stage: build
   <<: *build
   only:
     - schedules  # nightly build
     - master
+    - grdddj/tenv_tests
```
tsusanka, line 20: I would remove this as we have the `run/` prefix (see one line below).

ci/deploy.yml
```diff
@@ -0,0 +1,22 @@
...
+    - docker login $CI_REGISTRY -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD
+
+# Retagging the image in repository to be "latest"
+deploy image:
+  stage: deploy
+  <<: *deploy
+  only:
+    - schedules  # nightly build
+    - master
+    - grdddj/tenv_tests
```
tsusanka, line 17: dtto

tsusanka (approved):
In general looks good to me! 🎉 

What I am not sure about whether we do not want to finish https://github.com/trezor/trezor-user-env/pull/140 first and migrate everything to GHA first. And then merge this. What do you think @grdddj @vdovhanych?

ci/deploy.yml
```diff
@@ -0,0 +1,22 @@
...
+    - /^run\//
+  script:
+    - docker pull $CONTAINER_NAME:$CI_COMMIT_SHA
+    - docker tag $CONTAINER_NAME:$CI_COMMIT_SHA $CONTAINER_NAME:test
```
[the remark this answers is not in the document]
grdddj: It should definitely keep the `$CONTAINER_NAME:$CI_COMMIT_SHA`, as it is already stored in the docker repository from previous stage. We only download it, tag it and "upload" it, but the "upload" is really just adding another tag onto that existing image `$CONTAINER_NAME:$CI_COMMIT_SHA` - so it should not consume any extra resources

docker/compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
   trezor-user-env-unix:
     container_name: trezor-user-env.unix
-    image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env
+    image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:test
```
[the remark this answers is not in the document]
grdddj: Just for the testing purposes, so that I am not re-building and changing the production image. Will/should be deleted before merging :)

tsusanka:
Let's rebase this on top of master and move to Github? Feel free to squash into one commit if it helps @grdddj.

grdddj:
The "native" `controller` tests are running fine here, for example https://github.com/trezor/trezor-user-env/runs/5785749998?check_suite_focus=true

Not sure if we can port the `image_test.yml` here, but at least for the time being it can be run on `Gitlab`. In almost 100 % of cases the results would be identical, the `image_test` is just testing that `run-nix.sh` is fine

grdddj:
So that we can merge this already, without any fear of affecting production images, I have deleted all the steps touching the images and put everything in one commit.

vdovhanych (approved):
Looks good :) LGTM. I will remove GitLab stuff when I kill the project there as we should now have everything on GH Actions.

PR #129, opened 2021-12-13
feat(python): introducing Poetry

grdddj:
As discussed in https://github.com/trezor/trezor-user-env/pull/124#issuecomment-991697879, adding support for `Poetry`, which enables us to immediately install new `python` packages


tsusanka (approved):
LGTM 🎉

PR #132, opened 2021-12-15
fix(ci): set direct group dependency proxy prefix

vdovhanych:
This will allow dev account to retry the job and will fix the error with permissions introduced in the new version.


grdddj (approved):
Approving the changes

PR #133, opened 2021-12-16
chore(docker): pin to last working nix docker image

vdovhanych:
This fixes the issue with the latest docker image that was deployed by the nix community. It moves away from the alpine image and uses native nixos as a base.


grdddj (approved):
Looks good!

Issue #137, opened 2022-01-24
Create screenshotting functionality


grdddj:
As per conversation with @tsusanka and @mroz22, it would be nice to allow `tenv` to make screenshot from the current device screen and send it to the client - most probably through the websocket (maybe as a `base64` encoded data) - or in the worst case, client can take it with `docker cp`

grdddj:
Situation after https://github.com/trezor/trezor-user-env/pull/178 was merged:
- screenshots are supported only for `TT` and `TR`
- support for `T1` is not possible as the screenshots need special `debuglink` commands (through `trezorlib`) during _every_  new connection - which is something we cannot do for example with `Suite`

tsusanka:
Thanks. So I am closing for now!

PR #138, opened 2022-01-25
feat(emulator): emulator-get-screenshot command to return base64 encoded screen

grdddj:
Connected with https://github.com/trezor/trezor-user-env/issues/137

Allowing for clients to get the current screenshot of the device screen - encoded as a base64 string, so it can be transmitted through websocket in JSON format.

It is not the cleanest solution, as we need to save _all_ the screens, and when somebody requests current screenshot, we will return the latest one. That is because firmware currently does not have the functionality to save _just_ the current screen.

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -1,3 +1,4 @@
+import base64
 import os
 import stat
 import sys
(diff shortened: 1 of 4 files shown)


mroz22 (approved):
I very much like it. Looking forward to testing it.

src/emulator.py
```diff
@@ -172,6 +176,15 @@ def start(version: str, wipe: bool, output_to_logfile: bool = True) -> None:
 
     version_running = version
 
+    # Saving the screenshots on any screen-change, so we can send the
+    # current screen on demand
+    # Only applicable to TT, T1 is not capable of screenshotting
```
tsusanka, line 181: @matejcik is this true? 🤔

tsusanka (approved):
I am merging this so Martin can play with it. I am not sure if t1 does not support screenshots as well, I vaguely remember doing that.

PR #142, opened 2022-02-23
fix(emulator): enable experimental_features in emulator.apply_settings

szymonlesisz:
looks like its working now (tested locally)


grdddj (approved):
Good, the new `0.13.0` `trezorlib` indeed has the `experimental_features` field.

Approving

PR #145, opened 2022-03-02
feat: pass strenght and backup_type to device.reset

mroz22:
Does it look correct? I had trouble running trezor-user-env locally. Kept getting 'emulator process died" but I gues it was not connected to these changes. So this is not tested yet. will try again tomorrow.

### DIFF
--- a/src/controller.py
+++ b/src/controller.py
@@ -3,6 +3,7 @@ import json
 import traceback
 from copy import deepcopy
 
+from trezorlib import messages
 from websockets.exceptions import ConnectionClosedError, ConnectionClosedOK
 from websockets.server import serve
 
@@ -211,7 +212,10 @@ class ResponseGetter:
             emulator.apply_settings(**settings_fields)
             return {"response": f"Applied settings on emulator {settings_fields}"}
         elif self.command == "emulator-reset-device":
-            emulator.reset_device()
+            emulator.reset_device(
+                self.request_dict.get("backup_type", messages.BackupType.Bip39),
+                self.request_dict.get("strength"),
+            )
             return {"response": "Device reset"}
         elif self.command == "emulator-get-screenshot":
             screen_base_64 = emulator.get_current_screen()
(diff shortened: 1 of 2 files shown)


grdddj:
Seems OK. I am able to run this branch locally, click `Reset` in `dashboard` and fresh `emulator` will get initialized.

However, I did not test any extra `backup_type`s, apart from the default one - but in case of errors, emulator would tell us.

For the future, please try to run `make style` before pushing - so that we do not need to check red `CI` due to failed style-checks. Also because of this the static-code-checks are not run, which is a pity :)

mroz22:
ugh, did run `make style` but it created way more changes than expected. Even in files I did not touch at all. 21ff566c4ad6c40fa3699f83c9ae6596f2c10f4f

grdddj:
> ugh, did run `make style` but it created way more changes than expected

Could be because you were not in `poetry shell`, where we define specific versions of `black` and `isort` - the styling tools.

The usual way how to set up the `tenv` environment is to enter `nix-shell` first and then do `poetry install` and `poetry shell` - that will make sure the environment is up-to-date (`CI` is always doing this)

src/controller.py
```diff
@@ -211,7 +208,10 @@ def run_emulator_command(self) -> dict:
             emulator.apply_settings(**settings_fields)
             return {"response": f"Applied settings on emulator {settings_fields}"}
         elif self.command == "emulator-reset-device":
-            emulator.reset_device()
+            emulator.reset_device(
+                self.request_dict.get("backup_type", messages.BackupType.Bip39),
+                self.request_dict.get("strength", None),
```
grdddj, line 213: Just a small thing here, `None` is not necessary as a second argument, `dict.get("xx")` is implicitly returning `None` when the key does not exist there

grdddj:
Thanks for the changes. Do you want to merge it already, or will there be some more changes?

mroz22:
Yes, please merge :pray:

PR #148, opened 2022-03-17
feat(ci): add building test image on pull requests

vdovhanych:


grdddj (approved):
Looks good to me, I was able to pull and run the image from `ghcr.io/trezor/trezor-user-env:148`. (When running it the dashboard was not available, but I guess it is unrelated)

PR #151, opened 2022-03-27
Improve run.sh script

tsusanka:
- Pulls git changes if on master branch (and `--no-pull` is not provided)
- Adds `--help`
- Opens `localhost:9002` in a web browser
- Some small refactor

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,34 +1,75 @@
 #!/usr/bin/env bash
 set -e -o pipefail
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+CLEAR='\033[0m'
 
-if [ -z "$1" ]
-  then
-    TENV_REGTEST="trezor-user-env-regtest"
-elif [[ $1 == --no-regtest ]]; then
-    TENV_REGTEST=""
-fi
-
+TENV_REGTEST="trezor-user-env-regtest"
+DASHBOARD_URL="http://localhost:9002"
 SYSTEM=$(uname)
+XHOST_ADDRESS=""
+SERVICE_NAME=""
+OPEN=""
+PULL=1
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+function usage() {
+  if [ -n "$1" ]; then
+    echo -e "$1";
+  fi
+  echo "Usage: $0 [-r --no-regtest] [-p --no-pull]"
+  echo "  -r, --no-regtest      Run tenv only, no regtest"
+  echo "  -p, --no-pull         Do not pull changes, neither git nor docker. Automatically set to True when not on 'master'"
+  echo ""
+}
+
+# parse params
+while [[ "$#" > 0 ]]; do case $1 in
+  -r|--no-regtest) TENV_REGTEST=""; shift;;
+  -p|--no-pull) PULL=0; shift;;
+  -h|--help) usage; exit 0; shift;;
+  *) usage "${RED}Unknown parameter passed: $1${CLEAR}\n"; exit 1; shift;;
+esac; done
 
 if [[ $SYSTEM == Linux* ]]; then
-
-    echo -e "Setup xhost for video device output"
-    xhost +
-    echo -e "Downloading latest images"
-    docker-compose -f ./docker/compose.yml pull trezor-user-env-unix $TENV_REGTEST
-    echo -e "Starting trezor-user-env"
-    docker-compose -f ./docker/compose.yml up trezor-user-env-unix $TENV_REGTEST
+    SERVICE_NAME="trezor-user-env-unix"
+    OPEN="xdg-open"
 
 elif [[ $SYSTEM == Darwin* ]]; then
-
-    echo -e "Setup xhost for video device output"
-    xhost +127.0.0.1
-    echo -e "Downloading latest images"
-    docker-compose -f ./docker/compose.yml pull trezor-user-env-mac $TENV_REGTEST
-    echo -e "Starting trezor-user-env"
-    docker-compose -f ./docker/compose.yml up trezor-user-env-mac $TENV_REGTEST
+    XHOST_ADDRESS="127.0.0.1"
+    SERVICE_NAME="trezor-user-env-mac"
+    OPEN="open"
 
 else
-    echo -e "Not a supported system - $SYSTEM"
+    echo -e "${RED}Not a supported system: $SYSTEM${CLEAR}\n"
     exit 1
 fi
+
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -1,34 +1,71 @@
...
+
+echo -e "Setup xhost for video device output"
+xhost "+$XHOST_ADDRESS"
+
+if [[ $TENV_REGTEST == "trezor-user-env-regtest" ]]; then
+    echo -e "${GREEN}Starting trezor-user-env with Bitcoin regtest${CLEAR}\n"
+else
+    echo -e "${GREEN}Starting trezor-user-env${CLEAR}\n"
+fi
+(sleep 10 && $OPEN "http://localhost:9002")&
```
tsusanka, line 69: Not really sure if this is a good idea as it is not connected to the result of `docker-compose`. I am bit worried it will confuse users when something fails.

grdddj: Not sure either, but in most cases, it should ve alright. In the worst case, the browser will say `Unable to connect`. We could theoretically `ping` that URL and only open it in the browser if it suceeds

run.sh
```diff
@@ -1,34 +1,71 @@
...
-    echo -e "Downloading latest images"
-    docker-compose -f ./docker/compose.yml pull trezor-user-env-unix $TENV_REGTEST
-    echo -e "Starting trezor-user-env"
-    docker-compose -f ./docker/compose.yml up trezor-user-env-unix $TENV_REGTEST
+# parse params
+while [[ "$#" > 0 ]]; do case $1 in
+  -r|--no-regtest) TENV_REGTEST=""; shift;;
+  -p|--no-pull) PULL=0; shift;;
+  -h|--help) usage; shift;;
+  *) usage "${RED}Unknown parameter passed: $1${CLEAR}\n"; shift;;
```
grdddj, line 30: It could be good to `exit 1` if someone inputs bad parameter - one can overlook this message and have bad expectations about the spawned application

run.sh
grdddj: This worked for me - `curl http://localhost:9002 -s -o /dev/null && xdg-open http://localhost:9002`, only opening the dashboard if it runs properly

run.sh
```diff
@@ -1,34 +1,71 @@
...
+esac; done
 
-elif [[ $SYSTEM == Darwin* ]]; then
+if [[ $SYSTEM == Linux* ]]; then
+    SERVICE_NAME="trezor-user-env-unix"
+    OPEN="xdg-open"
```
grdddj, line 35: Seems the `$OPEN` is not defined globally unlike all others, so we may do it. It also opens a webpage, so we may call it `BROWSER_OPEN` to make it more clear

run.sh
tsusanka: I tried a similar approach with netstat and grep which did not work for me, not sure why. Thanks, let me try this one.

tsusanka: Seems like this works! 

`(while ! $(curl http://localhost:9002 -s -o /dev/null); do sleep 1; done; $OPEN "http://localhost:9002")&`

grdddj: RIght, I was about to comment with almost the same one which worked for me :D `while ! curl http://localhost:9002 -s -o /dev/null ; do sleep 0.5; done && xdg-open http://localhost:9002`

run.sh
tsusanka: You are right, that is inconsistent, not sure whether it is needed to define ti globally but why not. Regarding the name I have intentionally named it "OPEN" as it opens the default program for that given argument - if you would use it on .txt file it would open a text editor. So it seems correct to me :)

run.sh
tsusanka: That already happens in the usage function doesn't it? 🤔 But that being said for --help we should probably return 0 right? 🤔

run.sh
tsusanka: 8d3e758dbc0d7c16c32f0008c9424644f5625676

run.sh
tsusanka: 8d3e758dbc0d7c16c32f0008c9424644f5625676

grdddj: I did not know it really opens _anything_, so `OPEN` is fine in that case :)

run.sh
grdddj: That's correct, it returns 1 indirectly via `usage`. Help on its own probably should exit with 0, but probably with 1 in this case of bad argument :) But not a big deal

run.sh
grdddj: Nice, I verified it works fine in `run.sh`. I just think the fixup points to a bad commit, it should not be to `simplify README` :)

grdddj: Also, theoretically, we are using `http://localhost:9002` more than once, so it might be useful to create `DASHBOARD_URL` variable

vdovhanych:
When I run the script I do not get new images pulled, is that also because of not being on master? I feel the new docker images should be pulled regardless of the branch.

tsusanka:
@vdovhanych yes, there is a [warning](https://github.com/trezor/trezor-user-env/pull/151/files#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35dR51-R53). I do not have a strong preference so let me change it 👍 .

run.sh
tsusanka: > Nice, I verified it works fine in run.sh. I just think the fixup points to a bad commit, it should not be to simplify README :)

I have accidentaly added the README commit 🤦. Let me fix it.

tsusanka: > Also, theoretically, we are using http://localhost:9002 more than once, so it might be useful to create DASHBOARD_URL variable

👍 79a2a07fd90116fac44f8f313f1c80ee929e2b10

run.sh
tsusanka: 3b9f9cc

tsusanka:
I think all fixed, looks good now? Let's do it 🚀

vdovhanych (approved):

grdddj (approved):
All good!

Just a possible thing about pulling the latest images - as `tenv` is being updated very un-frequently, most of the time there will be no update in the `latest` image. I am not sure if we would pull that whole image again the next day (as new image was built from `master`), or it would somehow recognize we already have it? Not pulling unnecessarily could save a lot of time for users.

(Looking at `Gitlab`, the images being pushed from the same `master` have a different `sha256` hash)

tsusanka:
@grdddj right, good thinking though!

Issue #153, opened 2022-03-28
Try to improve caching in Dockerfile


grdddj:
It could be possible to achieve more docker-cache-reuse when building the `image` by moving the `RUN nix-shell --run "./src/binaries/firmware/bin/download.sh"` command closer to the end - so that changes in this layer do not invalidate further layers, which most of the time do not change.

tsusanka:
This is basically a one-liner isn't it? Let's try it? 💪

grdddj:
Looking at it a little more, the cache invalidation in `CI` happens already at `Step 6/17 : COPY ./ ./trezor-user-env`, so any changes after this step do not have any influence on it.

[This](https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/2249150106), [this](https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/2253655280) and [this](https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/2256424368) are `build` jobs from the same commit - `Checking out 8365163e as master...` - but after copying all the `tenv` code (that step 6 above), they produce a different layer hash (all 5 layers before it have `---> Using cache` identifier).

According to some sources - [here](https://stackoverflow.com/a/49957825) and [here](https://forums.docker.com/t/cant-find-why-my-docker-images-digest-are-different-when-building-from-the-exact-same-repo-version/95990/9), the issue is in different timestamps on the files - seems this is caused by us always cloning the repository from scratch on the `CI` server, not reusing existing files. 

I am not sure if this is even worth "fixing" by some script that would always unite the modification time of all files to some constant value. However, I will try for some while, as it has the potential of saving a lot of space and time to pull new images.

tsusanka:
Not worth it as discussed in https://github.com/trezor/trezor-user-env/pull/154. Good idea though!

PR #154, opened 2022-03-29
Grdddj/docker more cache reuse

grdddj:
Experiment `PR` trying to address https://github.com/trezor/trezor-user-env/issues/153

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,5 @@
 src/binaries/firmware/repo/
 src/binaries/trezord-go/repo/
-docker
-!docker/create_version_file.sh
 logs
 .git
 .dockerignore
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 src/binaries/firmware/repo/
 src/binaries/trezord-go/repo/
 docker
+!docker/create_version_file.sh
 logs
 .git
 .dockerignore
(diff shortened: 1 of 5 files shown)


vdovhanych:
Just FYI we will no longer be using GitLab for building the image I feel this is just lost time as we do not control the GitHub runner and thus caching is not really something we would see any improvements to it.

grdddj:
> Just FYI we will no longer be using GitLab for building the image I feel this is just lost time as we do not control the GitHub runner and thus caching is not really something we would see any improvements to it.

I feel the same, it is probably not worth spending time with. We will have to live with building, pushing, storing and pulling every single image/layer separately without reuse (apart from the base image).

tsusanka:
Yeah this seems too wild 😢. So closing? Thanks @grdddj though!

grdddj:
> Yeah this seems too wild cry. So closing? Thanks @grdddj though!

Sure, we can close it as an experiment, together with the issue probably :) 

One thing from this theoretically could be used, and that is the inclusion of `.git` directory in `.dockerignore`, as it has `88 MB` currently and will be only growing. We could save that space in every image.

(one thing that the omission of `.git` directory does is `fatal: not a git repository (or any of the parent directories): .git` error at the end of `nix-shell` installation - can be seen at https://gitlab.com/satoshilabs/trezor/trezor-user-env/-/jobs/2263134247 at line `751` - but it does not seem influencing anything)

tsusanka:
Seems nice to have to omit the .git directory. How big is the image btw? Is 88 significant part?

tsusanka:
It is so big I would not worry about it.. For now..

Issue #159, opened 2022-04-04
arm emulators and bridges with the same version name


vdovhanych:
Would it be possible to have the emulator and bridge indicate the same version as x86 one? A use case for this is suite integration tests where they call for emulator version `2.4.3` fe. but on arm its `2.4.3-arm` what do you think @grdddj

grdddj:
I guess it could be done just by saving the bridges/emulators _always_ with the same name in `src/binaries/firmware/bin/download.sh`, even if they are `arm` (as we know that the container can contain _either_ `arm` or `nonarm` artifacts, not both, so there will be no name clashes).

As for the implementation of above, there could be some clever bash command that would strip `"-arm"` from the filenames, or we would write some small python script for it.

There would need to be some small changes in `binaries.py`, where we hardcode the bridges for `Dashboard`. 

---

`OR` we would just modify our logic of finding the binaries - when request to spawn `2.4.3` arrives, we could check if it exists, and if not, look for `2.4.3-arm`.

There are multiple ways how to solve this, not sure now what is the best

vdovhanych:
@tsusanka and @grdddj what do you guys think would be the best solution? I don't have any favorites tbh.

grdddj:
I would say the cleanest is this:
- all clients will send just the version, so `2.4.3`, `2-master` etc.
- in `emulator.py` and `bridge.py` we check if we are on `arm` - the already existing `IS_ARM` variable - and if it is, we will just append `-arm` suffix to the version ... `2.4.3` becomes `2.4.3-arm` and the appropriate emulator is searched for in the filesystem

The benefit is that clients do not have to care about anything, the server/websocket/container is responsible for sending them all the valid binaries and then validating they requested a valid one.

vdovhanych:
I like this solution, it's not even that hard to just add it.

tsusanka:
Sounds great 👍.

grdddj:
Alright, let's do it :) @vdovhanych do you want to try implementing it?

vdovhanych:
Not sure i will manage, i have some idea but it didn't work for me :/

grdddj:
Closing as https://github.com/trezor/trezor-user-env/pull/160 is already merged

PR #160, opened 2022-04-07
feat(emu, bridge): add ARM suffix if on ARM and suffix not there

grdddj:
Connected with https://github.com/trezor/trezor-user-env/issues/159:
- even on `ARM`, allow request for "classic" emulator/bridge version, like `2.4.3` - and adding the `-arm`/`-arm64` suffix automatically if not there

### DIFF
--- a/src/binaries.py
+++ b/src/binaries.py
@@ -72,8 +72,8 @@ def sort_firmwares(version: str) -> Tuple[int, ...]:
 def explore_bridges() -> None:
     # Send only suitable bridges for ARM/non-ARM
     if IS_ARM:
-        BRIDGES.append("2.0.31-arm64")
-        BRIDGES.append("2.0.30-arm64")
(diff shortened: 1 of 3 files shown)


grdddj:
> Ha, a bit different approach but it seems like this one is working as expected, gj :)

I chose this as a path of least resistance and most freedom - so that clients are not restricted and can still send  `2.4.3-arm` from `Dashboard` or even from websocket if they want. It required minimum amount of changes :)

tsusanka (commented):
LGTM, but could we rename the Bridges binaries to `-arm` instead of `-arm64` so it is consistent? Because that is what we are using for FWs correct?

grdddj:
> could we rename the Bridges binaries to `-arm` instead of `-arm64` so it is consistent

That is a good idea, to unify it. I will do it in this PR as well

Issue #157, opened 2022-03-31
Include one non-debug emulator for screenshotting purposes


grdddj:
One of the use-cases of `tenv` is screenshotting of the emulator, however, the `debug` red square in the top right corner is annoying.

It would be nice to package one `non-debug` emulator, probably from the latest `master`.

tsusanka:
I do not think that is possible, we need to have debug link to take screenshots and do other instrumentation to work with it. Without debug mode this is not possible. Or am I missing smth?

The best way to get rid of the red square is to have a python script that replaces it with a neighborhood color :). I even wrote that once, but I am afraid it is gone 😢 .

vdovhanych:
Honestly, the easiest way is to just get rid of it in some photo edit tool. Not sure we need to do anything about it.

grdddj:
It turns out that when choosing `Save screenshots` checkbox before starting the emulator, the red square goes away in next screen. (Trezor does some magic at the background, and when it found out the screenshotting mode is on, it does not render the red square even in debug mode)

Actually, Josef Tetek found this when he wanted to do a screenshot without red. I verified it works, we could theoretically have `Screenshot mode` instead of  `Save screenshots`, so users wanting to do it know it can help them

grdddj:
> It would be nice to do this because I think this will be quite common request for mktg/content people.

Or theoretically... why don't we make it a default option to save screenshots and therefore not show the red square? It could save some questions and troubles. Small issue is that the red square appears on the very first screen and disappears only after some interaction with emulator - I will try to find some `screen_refresh` function so that it is red-square-free from the very start.

---

> I have realized i am a bit confused about the screenshoting feature in tenv now. The "Get screenshot" returns a base64 encoded image - I can see that in the log, correct? And that seems to be working regardless the "Save screenshots" checkbox. 

Good point, seems we could improve this a little bit. Actually when you click `Get screenshot`, you will get base64 of the **latest** saved screenshot. So for us, who run `tenv` locally, there will most probably be some "leftover screenshots" from previous sessions. Therefore we get some result even when running emulator without `Save screenshots`. (Solution could be to remove `Get screenshot`, hardly anyone uses it anyway - I think I created it there for testing purposes and then did not delete it.)

---

> So.. what is doing the "Save screenshots" checkbox then? It autosaves screenshots somewhere?

It is telling the starting emulator to save screenshots under `logs/screens`. Our reason to do it is to always have access to the latest screenshot, so we can send the current screen content.

[...]

PR #162, opened 2022-04-12
Grdddj/screenshots improvements

grdddj:
Loosely connected with https://github.com/trezor/trezor-user-env/issues/157:
- saving all the screenshots from emulators, when users want - each emulator session has its own directory
- mounting a docker volume to enable saving screenshots from the running container
- documentation and explanation in dashboard

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ termcolor = "^1.1.0"
 trezor = "^0.13.0"
 websockets = "^10.1"
 psutil = "^5.8.0"
+Pillow = "^9.1.0"
 
 [tool.poetry.dev-dependencies]
 black = "^21.12b0"
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -6,6 +6,8 @@ services:
     network_mode: "host"
     environment:
       - DISPLAY=:0
+    volumes:
+      - ./../logs/screens:/trezor-user-env/logs/screens
 
   trezor-user-env-mac:
     container_name: trezor-user-env.mac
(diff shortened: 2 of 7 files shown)


grdddj:
I felt it could be useful to have that automatic way to delete the red square - and storing it in repo means we will not lose it

grdddj:
Rebased on `master` and updating python dependencies

pyproject.toml
```diff
@@ -10,6 +10,7 @@ termcolor = "^1.1.0"
 trezor = "^0.13.0"
 websockets = "^10.1"
 psutil = "^5.8.0"
+Pillow = "^9.1.0"
```
tsusanka, line 13: Does it need to be capitalized?

grdddj:
> Just some nit-picks. Could you also make a screenshot of the dashboard? Just wondering what it looks like now.

What changed are just the dots-underlined checkboxes, and when you hover over them, you will see some additional details about what that checkbox does

image

pyproject.toml
grdddj: I think it is a capital by default. And it is the same in `firmware` - https://github.com/trezor/trezor-firmware/blob/master/pyproject.toml#L57

src/helpers.py
```diff
@@ -25,3 +26,31 @@
 def log(text: str, color: str = "blue") -> None:
     logging.info(text)
     print(colored(text, color))
+
+
+def replace_red_square(screenshot_location: str) -> None:
```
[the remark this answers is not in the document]
grdddj: Sure, changed in [31eaac9](https://github.com/trezor/trezor-user-env/pull/162/commits/31eaac99b3904f31054d602d1c66768637fdd4fe)

tsusanka (approved):
LGTM!

PR #167, opened 2022-05-02
chore(ci): remove gitlab-ci configuration after switch to gh actions

vdovhanych:
- chore(ci): remove gitlab-ci configuration after switch to gh actions
- feat(ci): separate tenv and regtest builds
- chore(ci): rename action and setup to run on trezor repository only

This pr removes Gitlab ci and updates GH action. Separates build of tenv and regtest to two actions. Sets up that actions will only run on trezor repository (mainly if someone forks this they will not be getting failed actions every night) and some changes to action names.


.github/workflows/build_and_push.yml
```diff
@@ -1,5 +1,4 @@
-name: Build and push images to registry
-
+name: Bild and push trezor-user-env image
```
grdddj, line 1: This `bild` is a new slang-word? :D Used here at least two times

.github/workflows/build_and_push.yml
```diff
@@ -10,6 +9,7 @@ on:
 jobs:
   build_images:
     runs-on: ubuntu-latest
+    if: ${{ github.repository_owner	== 'trezor' }}
```
grdddj, line 12: There seems to be a lot of whitespace at the left of `==`. Same in the file below

grdddj (commented):
utACK, just some stylistic nits

tsusanka:
Leaving this one up to you guys 👌 :).

.github/workflows/build_and_push.yml
vdovhanych: Thanks, I missed this somehow :laughing:

grdddj (commented):
Thanks for the changes. I still have some questions/clarifications about the migration:

- will `Gitlab` stop completely as soon as we merge this to `master`? There is still that automatic pipeline from `master` every 12 hours
- are we sure that no one is still using the `Gitlab` image?
- is the `GHA` setup to also build new images every day? I see `0 1 * * *` cron, which would be fine
- in this `PR` I cannot see the `Build and push test image` job - is the new `  pull_request:
    types: [opened, reopened]` working as expected?
- ii past `PR`s I think all the stages were starting at the same time - can we make it a gradual process, like in `Gitlab` - so that we will not `build and push` the test-image if the `validation checks` and `controller tests` failed?

vdovhanych:
> Thanks for the changes. I still have some questions/clarifications about the migration:
> 
> * will `Gitlab` stop completely as soon as we merge this to `master`? There is still that automatic pipeline from `master` every 12 hours
 - Yes after we merge this gitlab won't run anything anymore.
> * are we sure that no one is still using the `Gitlab` image?
 - all projects that are using trezor-user-env should be switched to github image already
> * is the `GHA` setup to also build new images every day? I see `0 1 * * *` cron, which would be fine
 - yes it builds every night
> * in this `PR` I cannot see the `Build and push test image` job - is the new `  pull_request: types: [opened, reopened]` working as expected?
 - you're right i will only run on opened and reopend and not every push, i will fix this.
> * ii past `PR`s I think all the stages were starting at the same time - can we make it a gradual process, like in `Gitlab` - so that we will not `build and push` the test-image if the `validation checks` and `controller tests` failed?
 - good idea i will implement this here

vdovhanych:
I set up a pipline-like workflow. @grdddj what do you think? In my opinion, it's indeed better this way.

.github/workflows/pull_request_ci.yml
```diff
@@ -0,0 +1,67 @@
...
+          nix_path: nixpkgs=channel:nixos-unstable
+
+      - name: Download emulator and bridge binaries
+        run: nix-shell --run "./src/binaries/firmware/bin/download.sh && ./src/binaries/trezord-go/bin/download.sh"
+      - name: "Install poetry"
+        run: nix-shell --run "poetry install"
+      - name: Run the tests
+        run: nix-shell --run "poetry run make test"
+  # Build and push test trezor-user-env image if all validation jobs pass
+  build_images:
```
grdddj, line 42: We can rename it to `build_test_images`, so it explicitly says it is `test` and not `latest`

grdddj:
> I set up a pipeline-like workflow. @grdddj what do you think? In my opinion, it's indeed better this way.

Nice, it really ran all three steps in a pipeline, in order which we specified. ACK that.

It may be unrelated, but the `build_images ` failed because of `unable to download 'https://github.com/NixOS/nixpkgs/archive/7fad01d9d5a3f82081c00fb57918d64145dc904c.tar.gz': HTTP error 500` - https://github.com/trezor/trezor-user-env/runs/6271216570?check_suite_focus=true#logs

.github/workflows/pull_request_ci.yml
grdddj: And it is actually not `images`, but just one `image` - so I suggest `build_test_image` or `build_tenv_test_image`. Also it should mention `push`, so what about `build_and_push_tenv_test_image`?

vdovhanych:
yeah, that looks like some internal GitHub issue 😨

grdddj:
> yeah, that looks like some internal GitHub issue fearful

Right now even the first job did not succeed - a nice test that indeed the whole pipeline is cancelled as soon as something fails

grdddj (approved):
LGTM!

PR #168, opened 2022-05-03
chore(ci): update to the latest stable nix and official nix image

vdovhanych:
Updates nixos to upstream docker image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/trezor/nix:latest
+FROM nixos/nix:2.8.0
 
 # "fake" dbus address to prevent errors
 # https://github.com/SeleniumHQ/docker-selenium/issues/87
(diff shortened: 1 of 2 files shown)


vdovhanych:
so it looks like we're finally able to use the official nix image again. @grdddj do you remember what jobs were failing after the nix image was updated? I think it was something related to connect in the firmware repo.

grdddj:
> so it looks like we're finally able to use the official nix image again. @grdddj do you remember what jobs were failing after the nix image was updated? I think it was something related to connect in the firmware repo.

Not remember anymore. `tenv` is not used at so many places, so `Connect` tests in `firmware` is a probable option

vdovhanych:
What about just shipping it and seeing if it breaks anything? I know it's kind of yolo, but I tested the image locally, and everything seems to be working fine. @tsusanka @grdddj

grdddj:
Agreed, let's try and see. Revert in the worst case :D

PR #172, opened 2022-05-05
feat(controller): add a websocket client for quick local testing

grdddj:
For local testing, it is good to have a quick way to call some websocket endpoint with parameters. It is much better than having to implement and call things in `dashboard`


vdovhanych:
Yeah, why not, ca you add just quick info in the docs regarding this?

tsusanka:
And maybe let's move it to a new folder tools/ or tests/ instead of src? Because it is not a "production" code right?

grdddj:
> Yeah, why not, ca you add just quick info in the docs regarding this?

>And maybe let's move it to a new folder tools/ or tests/ instead of src? Because it is not a "production" code right?

Both done in [127eba2](https://github.com/trezor/trezor-user-env/pull/172/commits/127eba249b0fa62f549e047b0cdc9a07a14145b6) - created `tools` directory with small non-production scripts (moved the `delete_red_square.py` there as well) and created `tools.md` page for their documentation

vdovhanych (approved):

tsusanka (approved):
Awesome!

Issue #169, opened 2022-05-04
Get current screen text from T1


grdddj:
@ondracja from `QA` has a requirement to get the current text from `T1` screen.

Use-case:
- during the recovery process, we want to know what word is on screen - the `Cypress` tests need to get this information

Possibilities:
- there could be some way to get that info from `debuglink.state()` object
- we _can_ get a screenshot of the current page, and then somehow extract the text

ondracja:
After a discussion with @mroz22, we belive that we should be able to achieve our use case on our side without any adjustment to the `trezor-user-env` side. I suggest closing this issue for the time being.

grdddj:
Sure, will close it for now. If you would decide to use the `T1` screenshots, please ping us so we can add this to `tenv`, it is not there yet but hopefully could be added quickly.

PR #173, opened 2022-05-05
feat(emu): add `emulator-get-debug-state` command

grdddj:
Connected with https://github.com/trezor/trezor-user-env/issues/169:
- allowing to get at least some basic information about what is on the emulator screen
- new `emulator-get-debug-state` websocket command, which will return an object with sometimes useful data

Use-case:
- during `T1` recovery, it contains for example `"recovery_fake_word": "dish"` when there is a fake word, or `"recovery_word_pos": 18` when Trezor asks for a real word
- for `TT`, the most useful is probably `"layout_lines": ["RecoveryHomescreen", "Select number of words", ""]`, which tells what is currently displayed on the screen

CC @ondracja @mroz22 - this most probably solves your issue, please try with a testing image (should be downloadable soon from `ghcr.io/trezor/trezor-user-env:173`)

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -114,10 +114,18 @@
   - **arguments**:
     - **use_shamir**: `bool` (optional) - whether to use Shamir backup (SLIP39) - default is False (BIP39 will be used)
 
-
 - **emulator-get-screenshot**
   - **action**: get current screen encoded as base64
   - **response**: `{"response": str}`
+
+- **emulator-get-debug-state**
+  - **action**: get current debug state, for example to check what is on the screen
+  - **response**: `{"response": dict}`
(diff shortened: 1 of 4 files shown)


grdddj:
Sorry for the force-pushing, I had an issue with testing the `emulator-get-screenshot`, there is indeed something strange on the background when setting `save_screenshots: True`, as @vdovhanych pointed out recently

vdovhanych (approved):

vdovhanych:
@grdddj I will merge it then?

grdddj:
> @grdddj I will merge it then?

Yes, we can do that. Thank you :)

vdovhanych (approved):

PR #174, opened 2022-05-09
fix: adding procps to nix-shell for checking emulator status by tenv

vdovhanych:
This fixes an issue with procps not present for checking emulator status in the new image.


grdddj (approved):
LGMT :)

Issue #176, opened 2022-05-24
Add support for Trezor R emulator


grdddj:
It would be nice to add `Trezor R` section, where people could try the current state and/or experiment with the new prototype layouts in recent PRs. In the future, we will need to support the `model R` anyway.

It could be also a good use-case for @sime, to have access to all the `model R` emulators in one place, without the need to always change the branch and rebuild the emulator (just pick the one from menu and start it).

grdddj:
Done in https://github.com/trezor/trezor-user-env/pull/178

PR #178, opened 2022-05-25
Support model R emulators

grdddj:
Connected with https://github.com/trezor/trezor-user-env/issues/176:
- supporting model R in `tenv`

Is a little hacky, for couple of reasons/limitations:
- we need the latest `trezorlib` for `model R` support, but it is unreleased
  - I am unable to install it from our github repo through `poetry install` ([known issue](https://github.com/python-poetry/poetry/issues/755)), so I am doing it manually through `pip` in `Dockerfile` as a last step
- we need functional `model R` emulator(s), but they are not yet released
  - in current `master` we do have it running, just with a very limited capabilities
  - therefore, I am hardcoding download of from the "most featured branch" currently ([grdddj/trezor_r_passphrase_input](https://github.com/trezor/trezor-firmware/tree/grdddj/trezor_r_passphrase_input)), which has prototypes of PIN entry, seed backup, seed recovery and passphrase entry

It is however easily possible for anybody to download any branch through the `Start from URL` feature - just giving it URL to some Gitlab job of `core unix frozen R debug build` (for example https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2498092302) and choosing `TR`.

What is currently missing is support for `ARM` - sorry guys @vdovhanych @sime @tsusanka - but it could be possible to add some `model R ARM build` info `firmware CI` :)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 src/binaries/firmware/repo/
 src/binaries/trezord-go/repo/
+docker/Dockerfile
+docker/compose.yml
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,8 +23,7 @@ RUN ls /
 RUN ls /trezor-user-env
 
 RUN nix-shell --run "./src/binaries/firmware/bin/download.sh"
-# patch emulator binaries for nix
-RUN nix-shell --run "autoPatchelf src/binaries/firmware/bin"
+RUN nix-shell --run "./patch_emulators.sh src/binaries/firmware/bin"
 RUN nix-shell --run "./src/binaries/trezord-go/bin/download.sh"
 
 RUN nix-shell --run "./docker/create_version_file.sh"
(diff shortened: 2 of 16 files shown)


grdddj:
> I will try to build a model R emulator on arm, if that works then there should be no reason it shouldn't work on m1. I will let you know here.

Thanks. I think the `arm` version should be fine already - `firmware CI` built it - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2510249260 - and @sime verified it can be run through `tenv`

tsusanka:
@grdddj is this ready to be reviewed and merged? Btw matejcik made a hotfix after the first trezorctl release (0.13.2), does this include that instead of the first version?

tsusanka:
Ok just ping me if this would wait for me please :).

grdddj:
> Ok just ping me if this would wait for me please :).

@tsusanka pinging you, as I think it is now ready for review and merge :)

The PR contains even some "non-modelR" things - see the list of commits

tsusanka (approved):
We could bikeshed some things but I do not think it is needed :). LGTM.

I will merge this on Wednesday due to the holidays. cc @mroz22 just so you know if you notice something odd on Suite side after the merge.

Issue #179, opened 2022-05-26
Emulators downloaded from URL cannot be started in Docker container


grdddj:
When running `tenv` through a docker container and using the `Start from URL` feature, the emulator will get correctly downloaded into `/trezor-user-env/src/binaries/firmware/bin/XXX`, but `tenv` will complain that `FileNotFoundError - [Errno 2] No such file or directory:`.

The file is visible in the filesystem under that exact name. However, when trying to execute it, even in the `shell` I am receiving `No such file or directory`. When trying to patch it for Nixos with `nix-shell -p autoPatchelfHook SDL2 SDL2_image --run "autoPatchelf trezor-emu-*"` it complains that ```/nix/store/nprym6lf8lzhp1irb42lb4vp8069l5rj-glibc-2.32-54/lib/libc.so.6: version `GLIBC_2.33' not found```.

I did not notice this problem earlier, but it is possible that this `Start from URL` never worked inside the Docker, as I am using just my local `tenv` mostly.

grdddj:
Fixed in https://github.com/trezor/trezor-user-env/pull/178

PR #180, opened 2022-05-26
fix(emu): set emulator to empty state when it fails to start

grdddj:
Loosely connected with https://github.com/trezor/trezor-user-env/issues/179:
- currently, when somebody uses `Start from URL` from `Docker`, it will fail to start the emulator, but also blocks any other (even valid) emulator to be started
  - (because when trying to stop the "nonrunning" emulator, it will cause an exception)
- solving it by setting `EMULATOR = None` in case it does not start, so we know it is not running (and will not try to stop it)

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -189,7 +189,12 @@ def start(
     if wipe and EMULATOR.storage.exists():
         EMULATOR.storage.unlink()
 
-    EMULATOR.start()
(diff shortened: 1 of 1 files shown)


grdddj:
Done as a part of https://github.com/trezor/trezor-user-env/pull/178

PR #186, opened 2022-07-01
Revert "fix(nix): correct autoPatchelf behavior"

vdovhanych:
This reverts commit 7740e3b.- Revert "fix(nix): correct autoPatchelf behavior"
- Revert "chore(ci): unite nix versions with firmware"
- chore: update trezorctl to latest version

After we merged these changes the tenv server wouldn't start up at all in docker. I have tested it and it was working for a bit but now its not even starting after the latest rebuild. Also updating trezorctl version to latest.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,8 +23,6 @@ RUN ls /
 RUN ls /trezor-user-env
 
 RUN nix-shell --run "./src/binaries/firmware/bin/download.sh"
-# patch emulator binaries for nix
-RUN nix-shell --run "autoPatchelf src/binaries/firmware/bin"
 RUN nix-shell --run "./src/binaries/trezord-go/bin/download.sh"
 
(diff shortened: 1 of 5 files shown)


vdovhanych:
Lets wait for testing image

vdovhanych:
I guess it's because it is used as the base docker image in there.

grdddj:
Artifacts from https://gitlab.com/satoshilabs/trezor/trezor-suite/-/jobs/2666971555 however show that the `tenv` was started - so the issue might be just `Suite` tests connecting to it.

It might have problems resolving the websocket endpoint maybe. That is the issue we had in `firmware` and that was solved by https://github.com/trezor/trezor-firmware/pull/2357 - in short, `CI` servers could not resolve `localhost`, so we had to change it to `127.0.0.1`. If `Suite` and `firmware` use the same `Gitlab` servers, that might easily be the case

vdovhanych:
closing this as we fixed it with docker update

PR #188, opened 2022-07-16
chore: remove docker form nix-shell

### LINKED ISSUE #175: speeding up nix-shell native trezor-user-env server
Would it make sense to remove docker, docker-compose and xhost from shell.nix to improve nix-shell loading when using the tenv natively with nix?

vdovhanych:
As there is no need for docker to be present in nix-shell, I would remove it and save us some space in the docker image.

closes #175


grdddj:
I am not sure this is possible, as then users would need to have `docker-compose` natively on their machines - currently, they do not have to, they just enter the `nix-shell`.

We can theoretically assume they have `docker` already, and so delete only that from `shell.nix`, leaving `docker-compose` there

grdddj:
Side-note - it seems like the tests are failing in `download.sh` script because we cannot download the `R` emulator from [Gitlab URL](https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2498092190/artifacts/raw/core/build/unix/trezor-emu-core) - and that would probably cause issues when building the image as well.

I created an issue for that - https://github.com/trezor/trezor-user-env/issues/189, will create a `PR` for that soon

vdovhanych:
That's something I thought. It should be looked on as a dependency of the system and not use nix-shell for that as we run it inside our docker container, docker in there is pointless for us, and it just takes up space. Yep, the test failed due to the inability to download the emulator.

grdddj:
> That's something I thought. It should be looked on as a dependency of the system and not use nix-shell for that as we run it inside our docker container, docker in there is pointless for us, and it just takes up space

I am not against it, as long as the users will be fine :)

vdovhanych:
removing the `docker` and `docker-compose` from nix-shell we save about 400MB on our image.
Screenshot 2022-07-18 at 15 15 18

grdddj:
> removing the `docker` and `docker-compose` from nix-shell we save about 400MB on our image.

That is very nice!

What is strange is that the `MAC`-images seem to be much smaller than the `Linux` ones. The `latest` in your case has `2.83GB`, my `latest` is `3.82GB`

```
ghcr.io/trezor/trezor-user-env    latest    f9da0f40194e   5 days ago   3.82GB
```

tsusanka (approved):
I am fine with this 👍.

vdovhanych:
Ok, merging this, docs are OK as it's stated there you need docker for this to work. We can do the firmware deletion later.

PR #195, opened 2022-08-04
Support for physical Trezor

grdddj:
Making `tenv` compatible with a physical `Trezor`, so it can be used for some automated tasks/tests in debug mode.

The biggest difference is in the spawning of `bridge` - by default for emulators, it is run with `-u=false`, which disables listening on `USB`. For a physical device, we do not add any arguments to the `bridge`.

To avoid confusion for users, dashboard shows that `tenv` is expected to work with a physical device - in both the `bridge` and `emulator` sections:
(it does not actually check if the physical Trezor is connected, but we could add it)

image

Support for physical `Trezor` is enabled by `$PHYSICAL_TREZOR` env variable.

`run.sh` script has a new `-t` flag, that will set it inside the container.

### DIFF
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -6,6 +6,10 @@ services:
     network_mode: "host"
     environment:
       - DISPLAY=$DISPLAY
+      # NOTE: it is possible to combine docker container and hardware device,
+      # but only for Linux, and only if the Trezor bridge is running locally.
+      # Second option of running with HW device is running the whole `tenv` locally.
+      - PHYSICAL_TREZOR=$PHYSICAL_TREZOR
     volumes:
       - ./../logs/screens:/trezor-user-env/logs/screens
       - ./../src/binaries/firmware/bin/user_downloaded:/trezor-user-env/src/binaries/firmware/bin/user_downloaded
--- a/run.sh
+++ b/run.sh
@@ -11,6 +11,7 @@ XHOST_ADDRESS=""
 SERVICE_NAME=""
 OPEN=""
 PULL=1
+PHYSICAL_TREZOR=0
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
 function usage() {
@@ -20,6 +21,7 @@ function usage() {
   echo "Usage: $0 [-r --no-regtest] [-p --no-pull]"
   echo "  -r, --no-regtest      Run tenv only, no regtest"
   echo "  -p, --no-pull         Do not pull changes, neither git nor docker. Automatically set to True when not on 'master'"
+  echo "  -t, --trezor          Support for physical Trezor instead of an emulator. WARNING: only forks for Linux, not MacOS."
   echo ""
 }
 
@@ -27,6 +29,7 @@ function usage() {
 while [[ "$#" > 0 ]]; do case $1 in
   -r|--no-regtest) TENV_REGTEST=""; shift;;
   -p|--no-pull) PULL=0; shift;;
+  -t|--trezor) PHYSICAL_TREZOR=1; shift;;
   -h|--help) usage; exit 0; shift;;
   *) usage "${RED}Unknown parameter passed: $1${CLEAR}\n"; exit 1; shift;;
 esac; done
@@ -70,6 +73,12 @@ fi
 
 # open localhost:9002 in web browser after the controller is launched
 (while ! $(curl $DASHBOARD_URL -s -o /dev/null); do sleep 1; done; $OPEN $DASHBOARD_URL)&
+
+if [[ $PHYSICAL_TREZOR -eq 1 ]]; then
+    echo "Will support physical Trezor instead of emulator."
+    # this env variable is being passed to the docker environment
+    export PHYSICAL_TREZOR=1
+fi
 
 # launch trezor-user-env
 docker-compose -f ./docker/compose.yml up --force-recreate $SERVICE_NAME $TENV_REGTEST
--- a/src/bridge.py
+++ b/src/bridge.py
@@ -28,7 +28,13 @@ def is_running() -> bool:
 
 
 def get_status() -> dict:
-    return {"is_running": is_running(), "version": version_running}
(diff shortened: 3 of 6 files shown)


matejcik:
cc @GabrielKerekes

gabrielKerekes:
👋 Thank you for this! Running with `./run.sh -r -t` doesn't seem to work for me though. Setting `PHYSICAL_TREZOR=1` and then running via `poetry run python src/main.py` works nicely though 👍 🤔 I've tested running via Docker on both a M1 mac and on an Ubunutu machine - both don't work. Do you have any idea what might be wrong? I have only checked out this PR and tried running `./run.sh -r -t` - I haven't done anything else so maybe I'm missing something here. Running with an emulator works.

grdddj:
> Running with `./run.sh -r -t` doesn't seem to work for me though
> I have only checked out this PR and tried running `./run.sh -r -t` 

This means you are downloading and running the `latest` image, that does not yet have this change. To use an image built on this branch, just append the image tag with the number of this `PR`.
 
So in `docker/compose.yml` replace the image `ghcr.io/trezor/trezor-user-env` by `ghcr.io/trezor/trezor-user-env:195` for your platform (you are on MAC, so in `trezor-user-env-mac` service)

gabrielKerekes:
Hm, now I'm getting a libusb error on both machines when I try to start the bridge.
image
I tried Googling around, but it led me nowhere. Should I start the bridge via the web interface or should it be running locally (not in Docker)?

grdddj:
> Hm, now I'm getting a libusb error on both machines when I try to start the bridge.
> image
> I tried Googling around, but it led me nowhere. Should I start the bridge via the web interface or should it be running locally (not in Docker)?

I am getting the same error when starting the bridge from the container. It  is possible that the `libusb` is not well supported in docker or we are missing it there. Will try to fix it.

It should be independent of from where the bridge is running, for me it works even with a local one.

grdddj:
So far I was unable to make the `USB` work inside `docker`.

@gabrielKerekes  are you fine with the current situation of having two options?
 **a)** to run `tenv` in `docker` and the `bridge` locally
 **b)** to run `tenv` locally without `docker`

Option **b)** also gives you option to make some quick adjustments according to you use-case. And if I may ask, what is your use-case of running `tenv` with a hardware device? We might also find it interesting for some automated testing.

vdovhanych:
For usb to work inside a docker container you either have to run it in privileged mode(not recommended) or map the usb device inside that container with a device flag, for example `--device=/dev/ttyUSB0`

grdddj:
> For usb to work inside a docker container you either have to run it in privileged mode(not recommended) or map the usb device inside that container with a device flag, for example `--device=/dev/ttyUSB0`

I tried to put some of these things in the `compose.yml`, but none of those helped me.

```
devices:
      - /dev/ttyACM0:/dev/ttyUSB0
      - /dev/ttyUSB0:/dev/ttyUSB0
      ...
 ```

**Comment by vdovhanych** [comment]
I have tried a couple of things on mac and Linux, but it seems that on mac, it's not likely at all, and as for Linux, you need to run it in privileged mode with the volume mapping of the device. I was able to have USB in container using this docker run
`docker run -t -i --privileged -v /dev/bus/usb:/dev/bus/usb` but not on mac.

**Comment by grdddj** [comment]
Thanks for the analysis. Theoretically, we can support it only for local usage by removing the current `-t`flag for `run.sh`. Locally, users could manually do `PHYSICAL_TREZOR=1 python src/main.py`.

@gabrielKerekes would this be alright for you, so it can be merged?

**Comment by gabrielKerekes** [comment]
Sorry for not responding, I'm currently busy with other work. I'll try to take a look in the afternoon. As far as a I remember, I wasn't able to get option **a** to work i.e. local bridge didn't work with docker for me. Running with Docker would be ideal but I think it only working on Linux shouldn't be a problem for testing purposes.

> And if I may ask, what is your use-case of running tenv with a hardware device?
The use-case is to run end to end tests of a third-party app (e.g. Daedalus) with a physical device. My simple understanding is that the tests would automate a UI flow in the third-party app and in-between make calls to Trezor to e.g. confirm a transaction or public key export if needed.

**Comment by vdovhanych** [comment]
We were already messing with the USB in a docker container for our firmware repository, but it wasn't working correctly, and we had only problems with it. In the end, we switched it to a basic shell executor and omitted docker for the hw device tests.

**Comment by gabrielKerekes** [comment]
So I took a look and indeed the option **a** doesn't work on macOS for me - somehow the docker container doesn't communicate with the bridge. I tried playing around with the docker ports, but I wasn't able to figure it out. It works without problems on Linux. I think for testing purposes Linux is enough so this PR resolves the issue as is 👍

File: docker/compose.yml [hunk 1]
```diff
@@ -20,6 +21,7 @@ services:
       - "127.0.0.1:21325:21326" # macos needs proxy to override the "origin" of the trezord request
     environment:
       - DISPLAY=docker.for.mac.host.internal:0
+      - PHYSICAL_TREZOR=$PHYSICAL_TREZOR
```
vdovhanych, line 24: Let's get rid of it here as it's a no go on mac.

vdovhanych (changes requested):

vdovhanych:
Is it working for you then, Let's merge it in the current state?

grdddj:
> Is it working for you then, Let's merge it in the current state?

I am for it, it can be useful to have these possibilities. I will force-push with deleting the `PHYSICAL_TREZOR=$PHYSICAL_TREZOR` line for `MAC`

grdddj:
I have also included some small comments, just in case. We could check if the container is running on `MAC` and throw some error, but maybe in the future, it will be supported. It is not a widely-used feature, so I think it is alright as is

docker/compose.yml
grdddj: Done in force-push

vdovhanych (approved):
ok then :)

PR #198, opened 2022-08-19
fix: test regtest controller over host.docker.internal

vdovhanych:
the use case for this: 
when a mac user wants to run cypress tests native this allows the cypress test to send commands to regtest container from a different container (tenv in this case) 

test (localhost) -> tenv (container) -> tenv to regtest on locahost (container 2) typically, this doesn't work, but when we use `http://host.docker.internal`, it should work.


grdddj (approved):
Looks good

PR #201, opened 2022-10-04
feat(bridge): add new 2.0.32 bridges

grdddj:
Adding support for new bridges. I could only test the `linux` bridge. @vdovhanych please try the `mac` one

image


vdovhanych:
It's working fine :) it just shows version `2.0.33` 
Screenshot 2022-10-04 at 13 11 49

Related to the suite issue, so let's hold on for the issue to be solved there and replace binaries with the proper ones once it's updated there.

vdovhanych (approved):

PR #226, opened 2023-10-24
Grdddj/tests and typing

grdddj:
Couple small improvements (also see commit messages):
- fix some broken tests
- add test for TR
- add some type hints
- fix `mypy` test
- replace removed `trezorlib` debug functionality with a different approach

### DIFF
--- a/.github/workflows/check-validation.yml
+++ b/.github/workflows/check-validation.yml
@@ -3,7 +3,7 @@ name: "[Check] validate trezor-user-env"
 on: [pull_request]
 
 jobs:
-  # Run code validattion
+  # Run code validation
   validation-check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


vdovhanych:
But take a look at some of the tests. It failed.

grdddj:
> But take a look at some of the tests. It failed.

Right, it looks like we cannot unzip the downloaded master emulators in the `GA` filesystem - https://github.com/trezor/trezor-user-env/actions/runs/6627572607/job/18002872098?pr=226#step:4:8637

I will try to set up the tests in `CI` against the latest production version, e.g. `2.6.1` currently - which should be there 

It would be nice to have also a direct link to all the `master` emulators, instead of needing to unzip the archive (which also contains a lot of useless things)

grdddj:
Still some issues with running bridge - https://github.com/trezor/trezor-user-env/actions/runs/6638165351/job/18034060215?pr=226

I have removed the controller tests from CI and leave them local-only ... I am not sure it ever worked on `Github Actions` - previously, we had it on `Gitlab` and it was fine there

PR #228, opened 2023-11-21
chore(binaries): change latest emulators name from master to main

grdddj:
Naming changes from `master` to `main`

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -26,7 +26,7 @@
 - **emulator-start**
   - **action**: start the specified version of emulator (and if one already runs, kills it)
   - **arguments**:
-    - **version**: `str` (1.9.4, 2.4.0., etc.) - default is the latest TT (`2-master`)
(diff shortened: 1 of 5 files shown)


vdovhanych:
We should inform the suite team before so they can change their test to reflect the change to main.

vdovhanych (approved):

grdddj:
> We should inform the suite team before so they can change their test to reflect the change to main.

Thanks for the reminder, I wrote them on Slack

vdovhanych:
We can merge this, as they are informed about the change and they need the new images to be able to change this, it will fail either way until both is fixed :)

grdddj:
> We can merge this, as they are informed about the change and they need the new images to be able to change this, it will fail either way until both is fixed :)

Right, I have pushed a backwards-compatible fix, so we can definitely merge it now and it will not break anything for `Suite`. 

Seems with the new commit Merging got blocked - will you please approve again?

vdovhanych (approved):
lgtm

PR #229, opened 2023-12-06
docs: add info about need for non-root docker

peter-sanderson:
- `./run.sh` will fail if docker is allowed in root mode only


grdddj (approved):
Thanks!

PR #234, opened 2024-04-03
T3T1 support, space saving, GH actions emulators

grdddj:
Support for T3T1 emulators

Decreasing the size of Docker image

Getting latest/main emulators from GH-actions

TODO: remove the temporary commit and load `GH_TOKEN` from env


grdddj:
In the end, getting the latest core emulators from `trezor.io` --- still depending on `Gitlab` for legacy and arm builds.

@vdovhanych I am already satisfied with this, can you please approve the small Dockerfile changes? It is just setting up `Poetry` garbage collection and moving all the `RUN` steps into one layer - which both make the image smaller.

I did quite a lot of experiments with decreasing the image size, with the idea of deleting as much nix-connected stuff as possible.
- the majority of almost 4GB image is taken by stuff from `/nix/store`, which is however hard to delete, as it is needed for the app to run
- I tried to get rid of it with e.g. `nix-collect-garbage -d`, `nix-store --gc` or `nix-store --optimize` - which helps in theory when run in the running container - but in the image phase, it deletes something, just to have it installed again when starting the application
- One possibility is to convert the Dockerfile to start with `ubuntu/debian` base image - I tried it, it is possible to get the size around 1GB - however, I was experiencing issues with running the latest emulators and with bridges ... but this I guess is the ultimate way of making the image more user-friendly regarding the size and startup time (especially for Suite CI - @Nodonisko)
- It is certainly possible to somehow decrease the size while keeping the `nix` base image - e.g. build the application in one build-stage and then copy just it into another runtime-stage ... this article seems to be the best I read, which however utilizes `flake.nix` instead of `shell.nix` --- https://mitchellh.com/writing/nix-with-dockerfiles

There are two upcoming tasks/TODOs:
- remove the hardcoded `trezor-emu-core-T3T1-v2-main-arm` emulator, when it will be available from Gitlab CI - from https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/artifacts/main/download?job=core%20unix%20frozen%20T3T1%20debug%20build%20arm
- replace custom `trezorlib` installation with official one after new release - @matejcik - the custom installation from `trezor-firmware` repo takes more than a gigabyte of space in the image

vdovhanych (approved):
If we do not use the download option from GitHub then its fine like this, but read my comment for that.

docker/alpine-minirootfs-3.15.0-x86_64.tar.gz
vdovhanych, line 1: Why the hell are we adding alpine minirootfs to the repository? Is it so that the layer is cached for build?

grdddj: I copied the `Dockerfile` exactly as Stick suggested - from `firmware` repo - and it is there ... https://github.com/trezor/trezor-firmware/blob/main/ci/Dockerfile

vdovhanych:
Im already working on simple alpine image without nix, should be third the size

Nodonisko:
@vdovhanych Nice thanks! Every saved MB here = saved few minutes of runtime in total across all our CI jobs

PR #236, opened 2024-04-09
Dashboard improvements

grdddj:
Improving the dashboard code significantly - use `vue.js` library instead of vanilla JS.

Also, small UI/UX improvements - better buttons, more log outputs, bar on top showing when the request is pending, etc.

TODO: make sure the starting emulator from URL works properly


grdddj:
@vdovhanych please take a quick look, it is ready to merge

It contains some dashboard UI/UX improvements, as well as backend fixes regarding custom emulators and ARM

vdovhanych (approved):

PR #239, opened 2024-04-15
feat: use native apline image to reduce size of the container image

vdovhanych:


grdddj:
Seems this can be closed, as it is done in https://github.com/trezor/trezor-user-env/pull/253

PR #242, opened 2024-06-05
Update README.md

jvaclavik:


tsusanka:
I do not think that this belongs here - it is a setup for docker in general, not particular to trezor-user-env. Also I am somewhat sure this is not the correct solution, did you try adding yourself into the `docker` group? See https://stackoverflow.com/questions/48957195/how-to-fix-docker-got-permission-denied-issue

vdovhanych:
Yes, this is not an issue of tenv, but rather an issue of incorrectly setting up the docker.

jvaclavik:
OK, up to you. I wanted to save some time of our future Linux newbies (and their buddies) because this docker setting is set incorrectly by default on Linux, or at least on some of them (confirmed from multiple Linux users).

vdovhanych:
Yes, but you should follow the instructions on installing Docker on Linux, not add instructions for setting up Docker here. That might change, and it's not related to this project.

PR #250, opened 2024-07-10
Enable mempoolfullrbf=1 for Regtest

peter-sanderson:
With implementation of https://github.com/trezor/trezor-suite/issues/13160 and https://github.com/trezor/trezor-suite/issues/13159 it would be useful to make this config consistent with it.

### DIFF
--- a/docker/bitcoin-regtest/entrypoint.sh
+++ b/docker/bitcoin-regtest/entrypoint.sh
@@ -5,7 +5,7 @@ set -e
 cd /opt/coins/nodes/bitcoin_regtest
 
 echo "Starting bitcoin regtest backend service"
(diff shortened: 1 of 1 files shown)


docker/bitcoin-regtest/blockchaincfg.json
```diff
@@ -16,5 +16,6 @@
     "slip44": 1,
     "mempool_workers": 8,
     "mempool_sub_workers": 2,
-    "block_addresses_to_keep": 300
+    "block_addresses_to_keep": 300,
+    "mempoolfullrbf": 1
```
peter-sanderson, line 20: This is absolutely blind shot at this point. I am still figuring our where to put this config...

vdovhanych (approved):

PR #253, opened 2024-08-02
use emu arm from github

vdovhanych:
- **chore: update model names to the internal model ids**
- **chore: remove old binaries and unused scripts**
- **feat: change docker configuration to native setup without nix-shell and download latest arm builds from data.trezor.io**


grdddj (approved):
The new image works fine for me, I can run all the emulators.

Feel free to merge

PR #255, opened 2024-09-05
fix: bridge proxy kill children as well

martykan:
I think the bridge proxy has an issue with left over processes, so they should be killed the same way as in bridge.

It's not a critical issue, but gets rid of some errors such as this:
Screenshot 2024-09-06 at 09 56 52


grdddj (approved):
LGTM

PR #261, opened 2024-09-23
Nodejs bridge

grdddj:
Experimental support for node.js bridge --- just specify `"node"` as the bridge version


mroz22 (approved):
looks ok, did not test.

PR #263, opened 2024-09-24
fix: make the nodejs bridge not-default

grdddj:
Should fix Suite e2e tests, putting node-bridge at the last place in the list


mroz22 (approved):
good for a fix, I would still add some logic that marks the default verbosly.

PR #271, opened 2024-10-09
fix: emulator.get_debug_state response

szymonlesisz:
fixing problem 1:
`get_debug_state` tries to serialize non buffer fields/keys (1 indent missing)
edit: after further testing this change can be tricky. 
DebugLinkState may contain some other filed types not only bytes (like int for example) but in the same time they might be complex and not serializable.
see thp state:

<DebugLinkState: {'mnemonic_secret': b'all all all all all all all all all all all all', 'passphrase_protection': False, 'reset_entropy': b'', 'mnemonic_type': <BackupType.Bip39: 0>, 'tokens': ['{', '"component"', ': ', '"SimplePage"', ', ', '"active_page"', ': ', '0', ', ', '"page_count"', ': ', '2', ', ', '"content"', ': ', '{', '"component"', ': ', '"AddressDetails"', ', ', '"qr_code"', ': ', '{', '"component"', ': ', '"Frame"', ', ', '"title"', ': ', '{', '"component"', ': ', '"Label"', ', ', '"text"', ': ', '"Scan QR code to pair"', '}', ', ', '"content"', ': ', '{', '"component"', ': ', '"Qr"', ', ', '"text"', ': ', '""', '}', ', ', '"button"', ': ', '{', '"component"', ': ', '"Button"', ', ', '"icon"', ': ', 'true', '}', '}', '}', '}'], 'thp_pairing_code_entry_code': 417041}>

fixing problem 2:
UI client assumes that `dataObject.response` is a string but in case of debug state it is an object

example: send JSON to server
```
{"type": "emulator-get-debug-state"}
```

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -909,15 +909,18 @@ def get_debug_state() -> Dict[str, Any]:
         debug_state_dict: Dict[str, Any] = {}
         for key in dir(debug_state):
             val = getattr(debug_state, key)
-            # Not interested in private attributes and non-JSON fields (bytes)
+            # Not interested in private or uppercase attributes
             if key.startswith("__") or key[0].isupper():
                 continue
+            # Not interested in methods
+            if callable(val):
+                continue
+            # Transforming bytes to string
             if isinstance(val, bytes):
                 try:
                     val = val.decode("utf-8")
                 except UnicodeDecodeError:
                     val = val.hex()
-
             debug_state_dict[key] = val
 
         return debug_state_dict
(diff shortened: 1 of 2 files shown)


src/emulator.py
```diff
@@ -917,8 +917,7 @@ def get_debug_state() -> Dict[str, Any]:
                     val = val.decode("utf-8")
                 except UnicodeDecodeError:
                     val = val.hex()
-
-            debug_state_dict[key] = val
+                debug_state_dict[key] = val
```
grdddj, line 920: This does not seem right, it means we will only save byte values

szymonlesisz: yes, as i mentioned in the description and also discussed [on slack](https://satoshilabs.slack.com/archives/CL1D61PQF/p1728639551657579), im aware that this isnt right but user-env crasches without it (tries to serialize complex field)

grdddj: I pushed a fixup commit into this branch/PR, can you please confirm it works fine for you? I can merge it in a positive case

szymonlesisz: works for me

grdddj (approved):
Ready to merge

PR #275, opened 2024-10-30
T3W1 support

grdddj:
Adding support for the new emulator model.

NOTE: needs to have a `trezorlib` installed from firmware repo, as the released version does not support it yet.


pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["SatoshiLabs <info@satoshilabs.com>"]
 [tool.poetry.dependencies]
 python = "^3.11"
 termcolor = "^1.1.0"
-trezor = "^0.13.9"
+trezor = { git = "https://github.com/trezor/trezor-firmware.git", branch = "main", subdirectory = "python" }
```
vdovhanych, line 10: Could we maybe do a beta release instead fo doing this?? It increases the size of the built image significantly.

grdddj: It is up to @matejcik to do the `trezorlib` release - I mentioned it to @Hannsek that it is/will be needed to do. I was checking the image size and now it seems to be exactly the same as with released version - could be that `poetry` improved the git-repo installation process

vdovhanych: Ok lets merge it

vdovhanych (approved):
lets merge it

Issue #276, opened 2024-11-05
Bitcoin-only option in emulator


enjojoy:
It would be useful to have a possibility to make an emulated device Bitcoin-only, or pick a Bitcoin-only firmware. Sometimes it's needed to test new features.

Issue #237, opened 2024-04-10
Allow running emulator with a build from a given branch


Hannsek:

grdddj:
Could be solved by uploading all the firmware emulators to data.trezor.io - PR https://github.com/trezor/trezor-firmware/pull/4358

PR #278, opened 2024-12-02
feat: support starting emulator from a custom firmware branch

grdddj:
It is possible to specify firmware branch and we will try to find and start the specified emulator model from that branch.

Emulators are downloaded from `https://data.trezor.io/dev/firmware/emu-branches` dir - connected with https://github.com/trezor/trezor-firmware/pull/4358

NOTE: emulators for `arm` are not being built, so this feature is `Linux`-only ... but as CI is running on `Linux`, everybody can use it there (for automated `Suite` tests)


grdddj:
Fixes https://github.com/trezor/trezor-user-env/issues/276 - for branches, it supports getting `-btconly` emulator instead of `-universal`

Fixes https://github.com/trezor/trezor-user-env/issues/237 - but not working on `ARM`

mroz22 (approved):

PR #284, opened 2025-01-20
fix: do not watch layout for legacy T1B1 emulator

grdddj:
Accounting for the difference between layout behavior in legacy and the new emulators


mroz22 (approved):
lets try!

PR #292, opened 2025-02-27
fix: download and patch fw binaries on nixos

szymonlesisz:
i had it stashed for a while now and used it only locally

im running user-env natively on nixos and having some issues [since the configuration changed](https://github.com/trezor/trezor-user-env/commit/4cafe457980ce3d17f6c603dfde61a92e7ed6461)

1. dowloading binaries produces copies instead of override existing file. Wtih this fix it will skip dowloading
2. patching binaries does not work, the process doesnt throw any error but patched binary cannot be started

### DIFF
--- a/src/binaries/firmware/bin/download.sh
+++ b/src/binaries/firmware/bin/download.sh
@@ -60,8 +60,14 @@ files=(
   "T3T1/trezor-emu-core-T3T1-v2.8.7${suffix}"
 )
 
-for file in "${files[@]}"; do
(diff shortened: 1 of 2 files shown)


szymonlesisz:
i believe that failing test is not related to my changes

grdddj:
> i believe that failing test is not related to my changes

Right, the failure is known, and it comes from `trezorlib` - I am attempting to fix it here - https://github.com/trezor/trezor-firmware/pull/4664

grdddj (approved):
Useful things, thanks!

PR #298, opened 2025-03-19
feat: Get features api

HajekOndrej:


grdddj (approved):
Thanks for this. I have fixed it a little and added serialization for the whole `Features` object, so we can return it all to the client. Suite tests can log the whole payload, or selectively choose just `revision` and maybe other attributes.

PR #310, opened 2025-05-12
chore: update ghcr io auth token to github actions default

vdovhanych:


grdddj (commented):
I have cherry-picked the commit in my quicker-building branch (only linux build), and it failed there - with the same 403 Forbidden ... https://github.com/trezor/trezor-user-env/actions/runs/14974868605/job/42064512237

Let's see how it will be successful in the current branch - in progress ... https://github.com/trezor/trezor-user-env/actions/runs/14974666880/job/42063822741

vdovhanych:
retry it, it should be working. Merging it.

PR #321, opened 2025-08-27
chores

mroz22:


docker/Dockerfile-debian
```diff
@@ -50,7 +50,7 @@ COPY ./poetry.lock /trezor-user-env/
 RUN poetry self add poetry-plugin-export
 RUN poetry export -f requirements.txt --output requirements.txt --without-hashes --without dev
 RUN pip install -r requirements.txt
-RUN pip install "git+https://github.com/trezor/trezor-firmware.git@022c7fe8bee9bd373c724bf88f618a75658de625#egg=trezor&subdirectory=python"
+RUN pip install "git+https://github.com/trezor/trezor-firmware.git@3daedcc7778daa5fd1242b378f68a9407110f5d8#egg=trezor&subdirectory=python"
```
mroz22, line 53: I basically need to force a new image so to achieve this I decided to update python trezor-version

grdddj:
Main `emulators` (not `trezorlib/trezorctl`) get into the image during `download.sh` script, which runs when image builds ... https://github.com/trezor/trezor-user-env/blob/51efcaf8e2ffd1fc518db7f10fabb7d540c8ccd1/src/binaries/firmware/bin/download.sh#L92-L106

It relies on correct emulators being uploaded on those `data.trezor.io` URLs, which I think goes there from nightly firmware `CI`. So if you experienced having couple days old emulators in Suite CI, this can be one potential reason - something broke in firmware CI and new emulators were not uploaded.

Some other things to check:
- whether new `tenv` (latest) images are created periodically every 24 hours: https://github.com/trezor/trezor-user-env/actions
- (low probability) Suite tests may cache the `tenv` image and not download the latest one

Possible confusion:
- latest version of `main` emulators is completely independent of `trezorlib` version/commit-hash

mroz22:
Soo anytime there is a new emulator download -> hash of the image should be affected right? Because if not, then in the last couple of days 0 new emulators must have been uploaded because there is no hash change

Issue #326, opened 2025-09-23
Adopt uv as a package manager


grdddj:
It should be quicker to install dependencies during build, and also could support the installation of custom trezorlib from firmware repo commit

PR #327, opened 2025-09-23
Uv replacing poetry

grdddj:
Fixes https://github.com/trezor/trezor-user-env/issues/326:
- migrates `poetry` installation and runtime to `uv`, which has these advantages:
  - the installation is quicker
  - it is possible to natively install `trezorlib` from firmware repo - which was not possible before (it needed the pip installation in dockerfile step)


grdddj:
@mroz22 The last commit here updates the trezorlib version - same as in your https://github.com/trezor/trezor-user-env/pull/321

vdovhanych (approved):

PR #338, opened 2025-12-05
feat: update image hash based on dowloaded content

mroz22:


vdovhanych:
After the changes to how we download firmware, this is now redundant and outdated.

PR #345, opened 2026-03-13
bind services to localhost and validate emulator download urls

YashejShah:
### network binding
the websocket controller (port 9001) and http dashboard (port 9002) were both bound to `0.0.0.0`, making them accessible to anyone on the network. the controller has no authentication, so on a shared wifi or cloud vm with a public ip, anyone could remotely trigger emulator downloads, wipe state, or shut down the service.

switched both to `127.0.0.1` so they're only reachable from the local machine.

### emulator url validation
`start_from_url()` downloads a binary from any url, marks it executable, and runs it. combined with the unauthenticated websocket, this was essentially remote code execution over the network.

added a domain allowlist — only `data.trezor.io`, `gitlab.com`, and `github.com` are accepted as download sources.

## testing
ran the dashboard and controller locally, confirmed they still work on localhost. tested that an invalid domain raises ValueError.

### DIFF
--- a/src/emulator.py
+++ b/src/emulator.py
@@ -9,6 +9,7 @@ import threading
 import time
 import urllib.request
 from contextlib import contextmanager
+from urllib.parse import urlparse
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


src/emulator.py
```diff
@@ -163,6 +164,12 @@ def start_from_url(
         raise RuntimeError(f"Unknown model {model}")
     emu_path = binaries.USER_DOWNLOADED_DIR / f"{model_identifier}{emu_name}"
 
+    # Validate that the download URL points to a trusted domain
+    parsed = urlparse(url)
+    allowed_domains = ["data.trezor.io", "gitlab.com", "github.com"]
```
vdovhanych, line 169: Thanks for the PR, but URL validation is not really a security feature anyway. If we add `gitlab.com` or `github.com` in there, you can basically do whatever anyway. Because of that fact, I would say it's not worth the bother.

vdovhanych (commented):
As for pinning it to the local bind address, if it works fine with the Docker setup lets do it.

src/emulator.py
YashejShah: Fair point on the URL validation — if the allowlist includes broad domains like github.com anyway, it doesn't really narrow the attack surface. I'll drop that part and keep just the bind address change.

vdovhanych:
I tried to run this but i realized that this can't work from inside of the docker conainer. We would need to bind it to the docker container network inside. Since we run docker and then expose the ports lets leave it to the firewall of the user/machine where this runs on.

PR #348, opened 2026-03-18
VNC

OriginalEveres:
Ideally we should ditch the xQuartz. (and the client rendering)

adding the VNC server and client allows us to render the device in the docker container - thus the behaviour will be same on every platform. There is a VNC server that is exposed on port 5900 (you can connect with any VNC client - even from mobile within your local network) and a noVNC browser client, so when you open a device within the emulator - it opens a new tab with VNC client connected to the VNC server

changes:
- added VNC server
- added websocket VNC client


```
Emulator binary → renders GUI via SDL2 → Xvfb (virtual framebuffer in memory)
                                                ↓                                                                                                                                                                       
                                           x11vnc (reads pixels from Xvfb)                                                                                                                                              
                                                ↓                         
                                           VNC protocol on port 5900                                                                                                                                                    
                                                ↓                                                                                                                                                                     
                                           websockify (TCP→WebSocket bridge)                                                                                                                                            
                                                ↓                                                                                                                                                                     
                                           noVNC in browser (renders pixels) 
```


credit to @Ondra-Zik for his similar idea

### DIFF
--- a/docker/compose.local.yml
+++ b/docker/compose.local.yml
@@ -1,0 +1,14 @@
+# Override file for local builds. Use with:
+#   docker compose -f compose.yml -f compose.local.yml up <service>
+services:
+  trezor-user-env-unix:
+    image: trezor-user-env:local
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile-debian
+
+  trezor-user-env-mac:
+    image: trezor-user-env:local
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile-debian
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -4,7 +4,6 @@ services:
     image: ghcr.io/trezor/trezor-user-env
     network_mode: "host"
     environment:
-      - DISPLAY=$DISPLAY
       # NOTE: it is possible to combine docker container and hardware device,
       # but only for Linux, and only if the Trezor bridge is running locally.
       # Second option of running with HW device is running the whole `tenv` locally.
@@ -25,7 +24,6 @@ services:
       - "5900:5900"   # VNC port (when using VNC instead of X11)
       - "6080:6080"   # noVNC web viewer port
     environment:
-      - DISPLAY=docker.for.mac.host.internal:0
       - REGTEST_RPC_URL=http://host.docker.internal:18021
       - MACOS=1  # Enable macOS-specific bridge configuration (excludes legacy 2.* bridges)
     volumes:
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -23,6 +23,7 @@ services:
       - "21325:21325" # legacy bridge port
       - "21328:21328" # new bridge port
       - "5900:5900"   # VNC port (when using VNC instead of X11)
+      - "6080:6080"   # noVNC web viewer port
     environment:
       - DISPLAY=docker.for.mac.host.internal:0
       - REGTEST_RPC_URL=http://host.docker.internal:18021
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -22,6 +22,7 @@ services:
       - "9002:9002"
       - "21325:21325" # legacy bridge port
       - "21328:21328" # new bridge port
+      - "5900:5900"   # VNC port (when using VNC instead of X11)
     environment:
       - DISPLAY=docker.for.mac.host.internal:0
       - REGTEST_RPC_URL=http://host.docker.internal:18021
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -21,7 +21,7 @@ services:
       - "9002:9002"
       - "21325:21325" # legacy bridge port
       - "21328:21328" # new bridge port
-      - "5900:5900"   # VNC port (when using VNC instead of X11)
+      - "15900:5900"  # VNC port (using 15900 to avoid conflicts with local VNC servers)
       - "6080:6080"   # noVNC web viewer port
     environment:
       - REGTEST_RPC_URL=http://host.docker.internal:18021
--- a/shell.nix
+++ b/shell.nix
@@ -4,12 +4,21 @@ let
   nixpkgsSha256 = "09nbk2q4w3v8x3v4r2y2s7v3nk8n4wqzxgykkp7na9bhijry2zla";
 
   nixpkgsUrl = "https://github.com/NixOS/nixpkgs/archive/${nixpkgsCommit}.tar.gz";
+
+  pkgs = import
+    (builtins.fetchTarball {
+      url = nixpkgsUrl;
+      sha256 = nixpkgsSha256;
+    }) { };
+
+  novnc = pkgs.fetchFromGitHub {
+    owner = "novnc";
+    repo = "noVNC";
+    rev = "v1.5.0";
+    sha256 = "sha256-3Q87bYsC824/8A85Kxdqlm+InuuR/D/HjVrYTJZfE9Y=";
+  };
 in
-with import
-  (builtins.fetchTarball {
-    url = nixpkgsUrl;
-    sha256 = nixpkgsSha256;
-  }) { };
+with pkgs;
 stdenv.mkDerivation {
   name = "trezor-user-env-controller";
   buildInputs = [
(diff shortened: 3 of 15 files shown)


docker/compose.local.yml
OriginalEveres: made this, so that a local development setup is possible

src/emulator.py
```diff
@@ -36,6 +36,12 @@
 MODEL_RUNNING: binaries.Model | None = None
 EMULATOR: CoreEmulator | LegacyEmulator | None = None
 
+# VNC support
+VNC_DISPLAY = ":99"
```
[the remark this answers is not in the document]
OriginalEveres: check now plz

src/dashboard/js/index.js
```diff
@@ -192,6 +194,18 @@ const app = createApp({
...
+
+            // Show inline VNC viewer when emulator starts with VNC enabled
+            if (
+                this.emulators.useVnc &&
+                "response" in dataObject &&
+                typeof dataObject.response === 'string' &&
+                dataObject.response.includes("EM_DEVICE_STARTED")
+            ) {
+                this.emulators.vncActive = true;
+                this.$nextTick(() => this.reloadVnc());
```
OriginalEveres: Vue.JS stuff to wait till the DOM finishes the updates

src/dashboard/js/index.js
```diff
@@ -192,6 +194,18 @@ const app = createApp({
             }
 
             this.logEvent(`Response received: ${event.data}`, color);
+            console.log(dataObject);
```
Lemonexe: delet dis

src/emulator.py
```diff
@@ -277,7 +283,16 @@ def version_model() -> str:
             f"killing the already running one - {VERSION_RUNNING}",
             "red",
         )
-        stop()
+        # Only stop the emulator process, not VNC — start_display() will
+        # restart VNC cleanly if use_vnc is set, and stop() would cause
+        # port conflicts when VNC is immediately restarted.
+        assert EMULATOR.process is not None
```
Lemonexe: The `assert` is duplicate, see condition on L280

💭  I see that `assert` is used commonly in this python codebase – that's strange, assert is a tool for debugging and unit tests, not production code, but idk, I'm only a mediocre pythonist so maybe I'm missing something 🤷

src/emulator.py
```diff
@@ -322,6 +343,9 @@ def version_model() -> str:
 
     # Verifying if the emulator is really running
     time.sleep(0.5)
+
+    if use_vnc:
+        vnc.start_capture()
```
Lemonexe: why not start capturing right away? It's just an idea to play with – might make the coldstart faster if VNC is already capturing before emulator starts.
But maybe it doesn't make sense.

Lemonexe (commented):
Looks good, but a lot of nits.
And legacy X11 does not work at all, so we need to decide: fix it or kill it..

src/dashboard/js/index.js
```diff
@@ -509,7 +526,24 @@ const app = createApp({
...
+            }
+        },
+        openVncPopup() {
+            const w = 350;
+            const h = 666;
+            const left = (screen.width - w) / 2;
+            const top = (screen.height - h) / 2;
+            const popup = window.open("about:blank", "novnc-viewer", `popup=yes,width=${w},height=${h},left=${left},top=${top}`);
+            if (popup) {
+                popup.location.href = "http://localhost:6080/vnc_embed.html";
```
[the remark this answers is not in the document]
OriginalEveres: not relevant

src/dashboard/index.html
```diff
@@ -273,6 +285,22 @@ <h4>{{details.header}} ({{ model }})</h4>
...
+                    <div>
+                        <button @click="reloadVnc();">Reload</button>
+                        <button @click="openVncPopup();">Open in popup</button>
+                    </div>
+                </div>
+                <iframe
+                    id="vnc-iframe"
+                    src="http://localhost:6080/vnc_embed.html"
+                    class="vnc-iframe"
+                ></iframe>
```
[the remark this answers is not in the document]
Lemonexe: I don't think this can ever be applicable. Feel free to resolve.

src/vnc.py
```diff
@@ -0,0 +1,166 @@
...
+                    geo[k] = v
+            if "WIDTH" in geo and "HEIGHT" in geo:
+                clip = f"{geo['WIDTH']}x{geo['HEIGHT']}+0+0"
+                _log(f"Emulator window detected: {clip}")
+                return clip
+        except Exception as e:
+            _log(f"Window detection failed: {e}")
+        time.sleep(0.5)
+
+    return None
```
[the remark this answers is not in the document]
OriginalEveres: it can. I have changed it

src/vnc.py
```diff
@@ -118,18 +96,18 @@ def start_display() -> None:
...
         _x11vnc_process = None
 
-    clip_arg = _find_window_clip()
+    window_id = _find_window_id()
 
     cmd = ["x11vnc", "-display", DISPLAY, "-forever", "-nopw", "-shared"]
-    if clip_arg:
-        cmd += ["-clip", clip_arg]
+    if window_id:
+        cmd += ["-id", window_id]
```
Lemonexe, line 138: much better 😍

src/dashboard/js/index.js
```diff
@@ -195,11 +195,7 @@ const app = createApp({
             this.logEvent(`Response received: ${event.data}`, color);
 
             // Show inline VNC viewer when emulator starts
-            if (
-                "response" in dataObject &&
-                typeof dataObject.response === 'string' &&
-                dataObject.response.includes("EM_DEVICE_STARTED")
-            ) {
+            if (dataObject.emulator_started) {
```
Lemonexe, line 198: much better, thanks 👍

Lemonexe (approved):
Very nice. There are only some nits remaining.

Btw only now I noticed that you can scale the window – that's actually a huge improement because the X11 windows were too big and not scalable 👍

docker/compose.yml
```diff
@@ -22,8 +21,9 @@ services:
       - "9002:9002"
       - "21325:21325" # legacy bridge port
       - "21328:21328" # new bridge port
+      - "5900:5900"   # VNC port (when using VNC instead of X11)
```
martykan: what if we used a less common port, in case the user is running VNC on their machine?

src/dashboard/js/index.js
```diff
@@ -193,6 +194,12 @@ const app = createApp({
 
             this.logEvent(`Response received: ${event.data}`, color);
 
+            // Show inline VNC viewer when emulator starts
```
martykan, line 197: With this approach I can't see the VNC viewer if the emulator is already running from before, eg. when I reload the page

src/dashboard/css/index.css
```diff
@@ -227,3 +227,26 @@ hr {
...
+
+/* VNC viewer */
+
+.vnc-viewer {
+    display: flex;
+    flex-direction: column;
+    flex-grow: 1;
+
+    border-radius: 8px;
+    min-height: 400px;
```
martykan: I would do `min-height: 600px`, this seems too small.

This is a bit offtopic, but also I feel like the grid layout could be somewhat improved, I get only 1 column for viewports <1600px, which could still fit 2 columns IMO

docker/compose.yml
OriginalEveres: good idea

OriginalEveres: updated

src/dashboard/css/index.css
OriginalEveres: Alright, I have updated it to 600px.

Regarding the grid layout I would do it in the follow up PR

martykan (approved):
Happy with everything from my side

shell.nix
```diff
@@ -15,7 +15,7 @@ let
     owner = "novnc";
     repo = "noVNC";
     rev = "v1.5.0";
-    sha256 = "sha256-JCGnFMHRn0twZ+MjOSLENjCFNi1VACpuMKxlNJAiMaA=";
+    sha256 = "sha256-3Q87bYsC824/8A85Kxdqlm+InuuR/D/HjVrYTJZfE9Y=";
```
OriginalEveres, line 18: not sure how often the hash should change, but I'll double check

vdovhanych (approved):
I think we can merge this. I'm happy with the feature.

shell.nix
OriginalEveres: I added a wrong hash initially

PR #354, opened 2026-03-26
chore: refactor firmware download process and add docker layer chaing improvements

vdovhanych:
- Introduced a new Python script `download.py` to handle firmware downloads, replacing the previous Bash script `download.sh`.
- Updated `.gitignore` to exclude the new script from being ignored.
- Modified Dockerfile to copy the new script and execute it for downloading firmware.
- Adjusted GitHub workflows to include build arguments for cache management during builds.
- Updated development documentation to reflect changes in the firmware download process.


martykan (approved):
Looks good, everything seems to work

PR #363, opened 2026-04-16
feat: redesign dashboard with Trezor Suite design language

Hannsek:
## Summary

Rewrite the dashboard at `localhost:9002` as a card-based, theme-aware UI mapped 1:1 to the controller spec (see `docs/dashboard-overview.md` on the source branch). The structure is a 3-row shell grid: a full-width top bar, a sidebar on the left paired with the live device canvas, and the event log spanning the bottom row.

- **Top bar** — Trezor wordmark + "Emulator" label on the left, small spinner while waiting for a response, action buttons on the right (Emulator setup / Start / Stop / theme toggle).
- **Sidebar "Control"** — main header mirrors the Commands card h3; "Services" and "Tools" sub-headers mirror `.cmd-sec h5` (uppercase, trailing line). Services: Emulator, Bridge, Tropic Square (conditional on Trezor Safe 7), Regtest. Tools: Server. Each item opens a flyout for configuration; the running Bridge and Tropic Square inline their current version under the label. Hovering "Services" reveals a dots legend (running / stopped / unknown). A Reconnect button appears here only when the WebSocket drops.
- **Canvas** — Device card with the live VNC iframe (Reload / Pop out actions), Commands card grouped as Simulate (shown only for button-input models) / Setup — seed / Setup — lifecycle / Verify & debug / N4W1 NFC (T3W1), and an Event log with filterable kinds.
- **Flyouts** — Emulator (model grid, firmware list, launch-option tiles, custom firmware disclosure with URL/branch segmented control), Bridge (version list with modern/legacy split, local Suite bridge disclosure with a 2-step instruction + copy-to-clipboard, status-page link), Tropic Square, Regtest (mine / send BTC with spec-accurate defaults), Server (raw JSON payload + ping / close-WS / exit actions).

## Design system

- **Typography** — TT Satoshi throughout, JetBrains Mono for logs and code blocks (served via `@font-face` / `local()` — no remote font imports).
- **Themes** — Light and dark, toggled from the top bar and persisted to `localStorage`. All log colours use CSS variables so they stay readable in both themes. The VNC iframe src carries a matching `background` URL param so the canvas padding area tracks the theme.
- **Controls** — Tile-style on/off toggles replace iOS-style switches for launch options, passphrase, log destinations, and BTC-only. Numeric Shamir dials are fused with the Read & confirm Shamir button into a single grouped unit.

## Sidebar + flyout UX

- **Flyout anchoring** — panels slide in from behind the sidebar's right edge (sidebar sits above the flyout in z-index, plus a pseudo-element covers the 16 px viewport-left gap), so they appear to emerge from the sidebar rather than crossing the viewport. Inset layout (top: 72 px, bottom: 16 px, border-radius: 12 px) so the panel reads as an island.
- **Active item** — the sidebar item that spawned the open flyout is highlighted with a `primary-soft` background and a 2 px left accent rail.
- **Scoped backdrop** — dims only the canvas; top bar and sidebar remain legible and clickable, so the user can switch to a different module without closing first.
- **Instant swap between flyouts** — switching from one open panel to another uses a transient `.no-fly-transition` class on `<html>` to suppress the slide on both outgoing and incoming panels, making it feel like an in-place content swap. First open and final close still animate.
- **Accessibility** — every flyout close button is labelled; the disabled Tropic Square sidebar item uses `aria-disabled` + `@click.prevent` instead of the `disabled` attribute so its data-tooltip reason stays discoverable.

## Infrastructure

- `docker/compose.yml` bind-mounts `src/dashboard/` and `src/vnc_embed.html` so local edits are served by the container without rebuilding the image.
- `src/main.py` auto-starts the default bridge (or the first available one) on server boot, so the dashboard comes up with host connectivity ready. Opt out with `TREZOR_USER_ENV_NO_BRIDGE_AUTOSTART=1` (the controller test fixture sets this so start/stop tests see a clean slate).
- `src/vnc.py` gains an (optional) `MODEL_CLIP: dict[str, str]` plus a `model` kwarg on `start_capture()`. When an entry is populated, x11vnc is started with `-clip WxH+X+Y` to crop the stream to that model's device shell. No per-model rectangles are defined yet, so all models use a full-window capture — no behaviour change until values are added.
- `src/emulator.py` forwards the running model into `vnc.start_capture()`.
- `src/vnc_embed.html` reads a `background` query parameter and applies it to `html`, `body`, `#screen`, and `rfb.background` so the VNC padding area adopts the host page theme. The value is validated against `#RGB` / `#RGBA` / `#RRGGBB` / `#RRGGBBAA`; anything else falls back to `#808080` so arbitrary CSS (e.g. `url(...)`) can't be injected.

## Resilience

- WebSocket auto-reconnect with exponential backoff and a visible Reconnect button.
- `sendMessage()` and `sendMessageOnBackground()` both guard on `readyState === WebSocket.OPEN`, so background checks don't throw `InvalidStateError` during reconnects.
- `selectedEmulatorModel` is validated against `emulators.versions` on startup; stale / hand-edited localStorage values fall back to `T3W1`.

## Commits

- `chore(docker): mount dashboard and vnc_embed for local dev`
- `feat(server): auto-start bridge on boot, scaffold per-model VNC clip, theme-aware embed background`
- `feat(dashboard): redesign with Trezor Suite design language`

## Test plan

- [ ] `./run.sh` starts the stack and opens `http://localhost:9002` with the new dashboard
- [ ] Bridge auto-starts on server boot; sidebar shows the running bridge version under "Bridge" within ~500 ms. Launching with `TREZOR_USER_ENV_NO_BRIDGE_AUTOSTART=1` skips the auto-start
- [ ] `make test` runs the controller tests to completion (they rely on the skip env var being set by the test fixture)
- [ ] Sidebar Connection row is hidden when WS is connected; a Reconnect button appears when the WS drops and restores the connection when clicked
- [ ] Open Emulator flyout → pick each of the 5 models → firmware dropdown is populated from the backend and the device card header reflects the running model
- [ ] Launch options (Animations, Wipe on start, Log to file, Screenshot mode) toggle visually and are respected on `emulator-start`
- [ ] Custom firmware disclosure — URL tab populates model via URL inference, branch tab sends `btc_only` flag correctly, download banner shows while a binary is being fetched
- [ ] Bridge flyout — modern and legacy 2.x versions are grouped; Local Suite bridge disclosure shows "detected" when mounted and a two-step instruction + working copy-to-clipboard otherwise; status-page link opens `0.0.0.0:21328/status/` only when bridge is running
- [ ] Tropic Square flyout appears only when T3W1 is selected or running; sidebar item shows its version label when running; disabled state is announced by screen readers via `aria-disabled`
- [ ] Regtest flyout — mine (count 1–10 000, default 1, reward address optional) and send (default 10 BTC) both hit the backend
- [ ] Server flyout — raw JSON payload with prefilled `{"type": "specify"}` sends correctly; ping / close WS / exit actions work
- [ ] Commands card — on touch models (T2T1, T3T1, T3W1) the Simulate section is hidden; on button models (T1B1, T3B1) Yes/No work. All 3 seed presets (all / academic / custom), Wipe / Reset / Shamir, read-and-confirm mnemonic, read-and-confirm Shamir (shares/threshold), Get features, Set backup state, Safety checks off
- [ ] N4W1 NFC tap/clear works (T3W1), tag ID dial clamps to 1–100
- [ ] Flyout anchoring — open from closed: panel emerges from behind the sidebar. Switching from one flyout to another: content swaps without a slide. Closing: panel retreats behind the sidebar. Sidebar items remain clickable while a panel is open
- [ ] Hover the "Services" sub-header in the sidebar → dots legend appears to its right
- [ ] Theme toggle flips dark ↔ light and persists across reloads; VNC background adopts the theme on reload
- [ ] Event log — filters are functional (All / Out / Ok / Err / Raw / System), Clear empties the list, "N new events" chip shows when scroll is offscreen, colours remain readable in both themes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/dashboard/js/index.js
+++ b/src/dashboard/js/index.js
@@ -1,6 +1,17 @@
 const websocketUrl = "ws://localhost:9001/";
 const backgroundCheckPeriodMs = 500;
 
+// Safe localStorage wrapper — file:// in some browsers throws SecurityError.
+const store = {
+    get: function(key, fallback) {
+        try { return localStorage.getItem(key) || (fallback !== undefined ? fallback : null); }
+        catch (e) { return fallback !== undefined ? fallback : null; }
+    },
+    set: function(key, value) {
+        try { localStorage.setItem(key, value); } catch (e) { /* ignore */ }
+    },
+};
+
 const { createApp } = Vue;
 
 const app = createApp({
@@ -14,35 +25,46 @@ const app = createApp({
                 status: "unknown",
                 statusColor: "black",
                 hasSuiteLocal: false,
+                runningVersion: null,
+                autoStart: store.get('userEnvBridgeAutoStart') === 'true',
             },
             tropic: {
                 outputToLogfile: true,
                 status: "unknown",
                 statusColor: "black",
+                runningVersion: null,
             },
+            selectedEmulatorModel: store.get('userEnvSelectedModel', 'T3W1'),
             emulators: {
                 versions: {
                     T1B1: {
                         header: "Trezor One",
+                        inputType: "Buttons",
                         versions: [],
                     },
                     T2T1: {
                         header: "Trezor T",
+                        inputType: "Touch",
                         versions: [],
                     },
                     T3B1: {
                         header: "Trezor Safe 3",
+                        inputType: "Buttons",
                         versions: [],
                     },
                     T3T1: {
                         header: "Trezor Safe 5",
+                        inputType: "Touch",
                         versions: [],
                     },
                     T3W1: {
                         header: "Trezor Safe 7",
+                        inputType: "Touch",
                         versions: [],
                     },
                 },
+                runningModel: null,
+                runningVersion: null,
                 wipeDevice: false,
                 screenshotMode: false,
                 animations: false,
@@ -61,6 +83,9 @@ const app = createApp({
                 btcOnly: false,
             },
             emulatorDownloadMessage: "",
+            customFirmwareSource: "url",
+            customFirmwareOpen: false,
+            suiteMountOpen: false,
             emulatorCommands: {
                 seed: "",
                 shamirShares: 3,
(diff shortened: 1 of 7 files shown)


vdovhanych:
It's very strange when you click on something on the left side to have the menu pop up all the way on the other side. Would be better to have the menu expand on the same side or out of the button you click so you dont have to move focus to the other side..

https://github.com/user-attachments/assets/b32fd9f7-f4d5-4c5b-8de8-51a48cd4c8f9

Hannsek:
Already fixed.

docker/compose.yml
```diff
@@ -12,6 +12,8 @@ services:
       - ./../trezor-suite:/trezor-user-env/trezor-suite
       - ./../logs/screens:/trezor-user-env/logs/screens
       - ./../src/binaries/firmware/bin/user_downloaded:/trezor-user-env/src/binaries/firmware/bin/user_downloaded
+      - ./../src/dashboard:/trezor-user-env/src/dashboard
```
vdovhanych: this needs to be removed before merge

docker/compose.yml
```diff
@@ -30,6 +32,8 @@ services:
...
```
vdovhanych: should be removed same as for the above ones

src/main.py
```diff
@@ -46,6 +46,26 @@ def log(text: str, color: str = LOG_COLOR) -> None:
     log(f"Working from {os.getcwd()}")
 
     binaries.explore(args)
+
+    # Auto-start a bridge so the dashboard comes up with host connectivity
+    # already wired. Tests (and anyone who wants a clean slate) can set
+    # TREZOR_USER_ENV_NO_BRIDGE_AUTOSTART=1 to skip this.
```
vdovhanych: I don't think this is good behavior. Why do we want to autostarg the bridge? I see potentional issue with the CI.

martykan: I agree, this was actually my suggestion, but I wouldn't do it like this. 
I would do it only at the UI layer based on remembered state in localstorage

vdovhanych (commented):
Just noticed one thing. Launching the ts7 emulator will also laucch the tropic emulator but stopping the emulator only stops ts7 emu and tropic keeps running. Not sure if its the scop of this pr but we should probably kill both when you stop emulator.

vdovhanych:
It would be nice if you could make the window with logs larger, or maybe add a pop-up window that lets you navigate the logs more easily.

src/vnc.py
```diff
@@ -20,6 +20,14 @@
 DISPLAY = ":42"
 NOVNC_URL = "http://localhost:6080/vnc_embed.html"
 
+# Optional per-model clip rectangle inside the emulator SDL window.
+# Format is x11vnc's -clip "WxH+X+Y", relative to the window when combined
+# with -id. No per-model clip rectangles are currently defined, so every
+# model uses a full-window capture. Populating an entry here would crop
+# the stream to that model's device shell, letting the iframe's
+# theme-coloured background replace the emulator's grey SDL padding.
+MODEL_CLIP: dict[str, str] = {}
```
martykan, line 29: As the comment says - No per-model clip rectangles are currently defined
Is there a plan to add them here? Or maybe this should for a separate follow up PR?

OriginalEveres: I'm not sure if this is a right approach. 

I assume we want to show exactly what the binaries output and the grey SDL padding is rendered from the binaries as well. 

There was a version before that did not have the padding, so the solution would be IMHO, not render / have a flag to not render, the padding.

src/dashboard/js/index.js
```diff
@@ -89,27 +114,109 @@ const app = createApp({
...
+            openFly: null,
+            theme: store.get('userEnvTheme', 'dark'),
+            copyFlash: null,
+            vncCacheBuster: Date.now(),
+            logFilter: 'all',
+            unseenLogs: 0,
+            wsReconnecting: false,
+            wsRetries: 0,
+            wsRetryTimer: null,
+            WS_MAX_RETRIES: 4,
```
OriginalEveres, line 126: nit: `WS_MAX_RETRIES` is being used a const it seems, but is located in Vue's reactive data object, we should move it elsewhere (global const var)

src/vnc.py
Hannsek: I want to leave this for follow up PR.

martykan (approved):

martykan:
YOLO

PR #371, opened 2026-04-28
docs: update readme to match actual Trezor Suite configuration

### LINKED ISSUE #370: Update readme for "Connecting to Trezor Suite"
In the README.md instructions for running Trezor suite locally seems to be outdated as interface is showing the following:

```markdown
### Connecting to Trezor Suite

Trezor Suite normally relies on it's own bundled Bridge. To use the Bridge from `trezor-user-env` instead, you need to disable the bundled Bridge in Suite's debug settings.

1. To enable debug settings, open Trezor Suite -> Settings and 5x click on the "Settings" title at the top of the page.
2. Go to the "Debug" tab, scroll down to "Transport clients" and enable "Bridge Transport".
3. If you want to use a real device again, make sure to re-enable the "Bridge server" option.
4. Don't forget to reload the app
```

Image

53gur0:
Closes: https://github.com/trezor/trezor-user-env/issues/370


53gur0:
Not needed, the option is only present in the desktop app

PR #380, opened 2026-05-15
Prevent uv from using packages newer than 30 days

### LINKED ISSUE #378: Prevent uv from using packages newer than 30 days
Implement https://github.com/trezor/trezor-firmware/pull/6919 in this repository, too.

Will be done after the original PR is merged.

M1nd3r:
- Limits exposure of the repo to supply chain compromise -- as long as you use uv.
 - If it is necessary to break this rule and install a "young" package, it is necessary to whitelist the package -- use `exclude-newer-package` (https://docs.astral.sh/uv/reference/settings/#exclude-newer-package).

Note: To be able to use relative duration with `uv`, it was necessary to update `nixpkgs` in `shell.nix`. This has a side effect - nix now uses python 3.12 instead of python 3.11.

Note2: This check was implemented in trezor-firmware repo in [PR #6919](https://github.com/trezor/trezor-firmware/pull/6919), in trezor-definitions here: https://github.com/trezor/definitions/pull/54, more repos to come...

### DIFF
--- a/docker/Dockerfile-debian
+++ b/docker/Dockerfile-debian
@@ -42,7 +42,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Install uv for package management
-COPY --from=ghcr.io/astral-sh/uv:0.8.20 /uv /uvx /bin/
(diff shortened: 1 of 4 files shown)


martykan (approved):
LGTM, [passing build](https://github.com/trezor/trezor-user-env/actions/runs/26169606042)

PR #384, opened 2026-05-27
fix: Keep Tropic running when emulator is stopped to keep its state

HajekOndrej:


src/controller.py
```diff
@@ -281,14 +281,7 @@ def run_emulator_command(self) -> "ResponseType":
...
             return {"response": response_text, "emulator_started": True}
         elif self.command == "emulator-stop":
-            # emulator.stop() clears MODEL_RUNNING, so read it first. If the
-            # T3W1 emulator auto-started the Tropic model server on launch
-            # (see emulator-start handler above), tear it down here too so
-            # stopping the emulator cleans up everything it spawned.
-            was_t3w1 = emulator.MODEL_RUNNING == "T3W1"
             emulator.stop()
-            if was_t3w1 and tropic_model.is_running():
-                tropic_model.stop()
```
HajekOndrej, line 291: The intent here was correct to fully clean up, but this also causes the tropic to lose important secrets and session data, when started again and it often is needed to restart the trezor device in tests. There is currently not readily available a better way to keep the session alive than to keep the process running.

Vere-Grey (approved):
consider documentin the problem by comment

mroz22 (approved):

PR #385, opened 2026-05-28
feat(dashboard): warn that legacy trezord-go bridge is unsupported

mroz22:
## Summary

- Adds an amber warning banner above the legacy 2.x (`trezord-go`) bridge entries in the Bridge flyout.
- Modern `node-bridge` remains the only version supported by Trezor Suite. Legacy bridges are kept available, but the UI now makes their deprecated status explicit.

## Preview

Legacy bridge warning — dark and light theme

## Rationale

We want to keep legacy `trezord-go` available for a while (e.g. testing against older Suite versions), but signal clearly that it should not be used for normal workflows. This PR is the short-term step.

The follow-up PR that removes `trezord-go` entirely is **#386** — merge this one first, then land #386 once we are comfortable retiring legacy completely.

## Test plan

- [ ] Open the dashboard, expand the Bridge flyout
- [ ] Verify the amber warning appears above the `2.0.33` / `2.0.32` entries
- [ ] Verify the warning is hidden when no legacy bridges are listed (e.g. on macOS)
- [ ] Verify dark/light themes both render the banner readably

### DIFF
--- a/src/dashboard/css/index.css
+++ b/src/dashboard/css/index.css
@@ -1685,6 +1685,28 @@ label[data-tooltip]:hover::before {
 }
 .vlist-divider::after { content: ""; flex: 1; height: 1px; background: var(--border); }
 
+.bridge-legacy-warning {
+    display: flex;
+    align-items: flex-start;
+    gap: 8px;
+    padding: 8px 10px;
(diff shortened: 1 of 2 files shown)


martykan (approved):
LGTM

PR #390, opened 2026-06-19
[WIP, SLOP] Align native Nix setup with Docker emulator prep

peter-sanderson:
## Summary

- Make `run-nix.sh` prepare native Nix runs similarly to Docker:
  - download release emulator binaries
  - download current main/nightly emulator binaries
  - prepare node-bridge
  - patch emulator binaries for Nix before starting the controller
- Limit release emulator downloads to the latest 5 release artifacts per model.
- Keep current main/nightly emulator per model available.
- Fix Nix execution issues in shell scripts:
  - avoid `/bin/bash` shebang assumptions
  - avoid nested `nix-shell` from `patch-bin.sh`
  - call `auto-patchelf` with explicit runtime library paths

## Verification

- `python -m py_compile src/binaries/firmware/bin/download.py`
- `RELEASE_MINOR_LIMIT=5 nix-shell --run "uv run python src/binaries/firmware/bin/download.py releases"`
- `nix-shell --run "uv run python src/binaries/firmware/bin/download.py nightly"`
- `nix-shell --run "./src/binaries/firmware/bin/patch-bin.sh ./src/binaries/firmware/bin"`
- Firmware scanner output confirmed:
  - `T1B1`: main + 5 releases
  - `T2T1`: main + 5 releases
  - `T3B1`: main + 5 releases
  - `T3T1`: main + 5 releases
  - `T3W1`: main + 4 releases currently available from source

## Notes

- T3W1 currently resolves to main + 4 release artifacts because only 4 release artifacts were available from the source.
- Existing local `src/mcp/uv.lock` changes were intentionally left out of this commit.

### DIFF
--- a/src/binaries/firmware/bin/patch-bin.sh
+++ b/src/binaries/firmware/bin/patch-bin.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 SYSTEM_ARCH=$(uname -m)
 echo "System architecture: $SYSTEM_ARCH"
@@ -8,7 +8,17 @@ BINARY_DIR="${1:-./}"
 
 # ./src/binaries/firmware/bin/patch-bin.sh ./src/binaries/firmware/bin/
 if [ -n "$IN_NIX_SHELL" ]; then
-    nix-shell --run "autoPatchelf ${BINARY_DIR}trezor-emu-*"
+    mapfile -t EMULATORS < <(find "$BINARY_DIR" -type f -name 'trezor-emu-*' -executable | sort)
+    if [ "${#EMULATORS[@]}" -eq 0 ]; then
+        echo "No emulator binaries found under $BINARY_DIR"
+        exit 0
+    fi
+    if [ -n "${TREZOR_USER_ENV_AUTO_PATCHELF_LIBS:-}" ]; then
+        IFS=: read -r -a NIX_LIB_DIRS <<< "$TREZOR_USER_ENV_AUTO_PATCHELF_LIBS"
+    else
+        mapfile -t NIX_LIB_DIRS < <(find /nix/store -maxdepth 2 -type d -name lib 2>/dev/null | sort)
+    fi
+    auto-patchelf --ignore-missing --paths "${EMULATORS[@]}" --libs "${NIX_LIB_DIRS[@]}"
     exit
 fi
 
--- a/run-nix.sh
+++ b/run-nix.sh
@@ -1,13 +1,26 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
+set -euo pipefail
 
 DIR=$(dirname "$0")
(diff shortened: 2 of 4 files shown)


src/binaries/firmware/bin/download.py
```diff
@@ -196,7 +190,7 @@ def main() -> None:
     bin_dir = Path(__file__).resolve().parent
 
     releases_json_url = os.environ.get("RELEASES_JSON_URL", DEFAULT_RELEASES_JSON_URL)
-    minor_limit = int(os.environ.get("RELEASE_MINOR_LIMIT", "6"))
+    minor_limit = int(os.environ.get("RELEASE_MINOR_LIMIT", "5"))
```
vdovhanych: why this change?

src/binaries/firmware/bin/download.py
```diff
@@ -18,10 +18,6 @@
 )
 SUPPORTED_MODELS = ["T1B1", "T2T1", "T3B1", "T3T1", "T3W1"]
 
-# supports one pinned release per model in addition to the 6 latest releases
-PINNED_RELEASES: dict[str, str] = {"T2T1": "2.3.0"}
```
vdovhanych, line 22: why have you removed pinned versions??

src/binaries/firmware/bin/download.py
```diff
@@ -90,14 +86,6 @@ def build_release_paths(
     for model in SUPPORTED_MODELS:
         family = "legacy" if model == "T1B1" else "core"
         model_paths: list[str] = []
-        pinned_version = PINNED_RELEASES.get(model)
```
vdovhanych, line 93: same as above

src/binaries/firmware/bin/download.py
```diff
@@ -144,7 +130,15 @@ def download_releases(
...
 ) -> None:
     releases_json = fetch_json(releases_json_url)
-    for file_path in build_release_paths(releases_json, suffix, minor_limit):
+    selected_paths = build_release_paths(releases_json, suffix, minor_limit)
+    selected_names = {Path(file_path).name for file_path in selected_paths}
+    for emulator in bin_dir.glob("trezor-emu-*"):
+        if "main" in emulator.name or "url" in emulator.name:
+            continue
+        if emulator.name not in selected_names:
+            emulator.unlink()
```
vdovhanych: I don't understand why this is needed?

src/binaries/firmware/bin/patch-bin.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
vdovhanych, line 1: keep in mind this is executed in docker as well if you tested it and it doesn't break the tenv docker images then ok.

vdovhanych (commented):
I think this will break the tenv in Docker if we merge it.

PR #393, opened 2026-07-14
feat(dashboard): install custom firmware from an uploaded binary file

OriginalEveres:
## What

Adds a **"From file"** option to the Custom firmware panel in the dashboard, alongside the existing **From URL** and **From branch** flows. You can now upload an emulator build directly instead of pointing at a URL or branch.

## How it works

- The selected file is base64-encoded in the browser and sent over the existing websocket as a new `emulator-start-from-file` command.
- The backend decodes it, writes it to `user_downloaded/`, makes it executable + patches it for Nix, registers it, and launches it — reusing the same tail as `start_from_url`/`start_from_branch`.

## Notable details

- **Validation:** rejects non-ELF artifacts early and names the detected format (e.g. a `TRZV` device firmware `.bin`, a gzip/zip archive, or an HTML error page), instead of failing later with a cryptic `Exec format error`.
- **Always wipes:** a re-upload reusing the same name overwrites the previous binary, so the profile is always wiped to avoid running a stale profile against a different build.
- **Websocket max message size** raised to 256 MiB so firmware binaries fit (default was 1 MiB).
- **Log hygiene:** the base64 blob is redacted from both the client and server request logs.
- **Docs:** `docs/controller.md` documents the new command.

## Testing

- `py_compile` / `node --check` pass.
- Manually exercised the upload flow in the local Docker dashboard; confirmed the validation path (rejecting a `TRZV` firmware image and a non-ELF file with clear messages).
- Not yet verified end-to-end with a real ARM emu build launching — draft pending that check.

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -52,6 +52,17 @@
     - **output_to_logfile**: `bool` (default=True) whether the debug output should go to a logfile - otherwise it goes to stdout
     - **save_screenshots**: `bool` (default=False) whether to save screenshots to enable calling **emulator-get-screenshot**
 
+- **emulator-start-from-file**
+  - **action**: saves an uploaded emulator binary and runs it
+  - **note**: the profile is always wiped for an uploaded binary, since reusing a name overwrites the previous binary and a stale profile must not be run against a different build
+  - **arguments**:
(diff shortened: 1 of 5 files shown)


OriginalEveres:
@coderabbitai review

PR #395, opened 2026-07-16
feat(bootloader): mock device that presents as being in bootloader mode

OriginalEveres:
## Why

The Trezor emulator has **no bootloader** by design — the [docs](https://docs.trezor.io/trezor-firmware/core/emulator/index.html) state "neither boardloader nor bootloader and no firmware uploads." So you cannot put the emulator into bootloader mode (rebooting the firmware emu just kills it; there is no bootloader emulator artifact anywhere).

But Suite/host apps don't need a *real* bootloader — they only need the device to **present** as being in bootloader mode (`Features.bootloader_mode = true`) to render and drive their bootloader / firmware-update flows. This PR provides exactly that.

## What

A small **mock device** (`src/bootloader_mock.py`): a UDP server that speaks Trezor wire **protocol v1** (plaintext) on the emulator port and reports `bootloader_mode = true`.

Key insight that makes it simple: the **bootloader always uses protocol v1 on every model** (THP is firmware-only — confirmed: trezorlib's THP layer has zero bootloader awareness, the v1 layer handles `bootloader_mode`). So a single v1 mock covers **both T2T1 and T3W1**, with no Noise/THP handshake to implement.

It answers `Initialize`/`GetFeatures`, `Ping`, `Cancel`, `WipeDevice`, and drives a stubbed firmware-update flow (`FirmwareErase → FirmwareRequest×N → Success`) so the whole Suite update flow can run to completion (as a no-op flash).

- **controller:** `emulator-start-bootloader` command (T2T1/T3W1 only; replaces any running emulator since it takes the emulator UDP port); `emulator-stop` also stops the mock.
- **dashboard:** a "Bootloader mode (mock)" section in the emulator flyout with a model picker + start button.
- **docs:** documents the command.

## Verified

- ✅ **Protocol correctness** via trezorlib over direct UDP: liveness (`PINGPING`/`PONGPONG`), protocol `probe()` → v1, `GetFeatures` → `bootloader_mode=True` / `internal_model=T3W1`, and a **full `firmware.update()` runs to completion** with progress callbacks.
- ✅ **Node bridge discovery**: the node bridge (that Suite talks to) enumerates the mock and acquires a session.
- ✅ **End-to-end with Suite**: with the mock running on the emulator port, **Suite reports "Device connected in bootloader mode"** (T3W1). 🎉

### DIFF
--- a/docs/controller.md
+++ b/docs/controller.md
@@ -52,8 +52,13 @@
     - **output_to_logfile**: `bool` (default=True) whether the debug output should go to a logfile - otherwise it goes to stdout
     - **save_screenshots**: `bool` (default=False) whether to save screenshots to enable calling **emulator-get-screenshot**
 
+- **emulator-start-bootloader**
+  - **action**: start a mock device that presents as being in bootloader mode (the real emulator has no bootloader), so Suite/host apps can test bootloader & firmware-update flows. Speaks Trezor wire protocol v1 over the emulator UDP port; replaces any running emulator (no VNC display).
+  - **arguments**:
+    - **model**: `str` - only `T2T1` and `T3W1` are supported
+
 - **emulator-stop**
-  - **action**: stop the emulator
+  - **action**: stop the emulator (and the bootloader mock, if running)
 
 - **emulator-setup**
   - **action**: perform the emulator setup
--- a/src/bootloader_mock.py
+++ b/src/bootloader_mock.py
@@ -1,0 +1,250 @@
+"""A minimal mock Trezor device that reports it is in *bootloader mode*.
+
+The real Trezor emulator has no bootloader (see docs: "neither boardloader nor
+bootloader and no firmware uploads"), so there is no way to put the emulator
+into bootloader mode. Suite / host apps only need the device to *present* as a
+bootloader though - i.e. answer `Initialize`/`GetFeatures` with
+`Features.bootloader_mode = True` over the wire - to exercise their bootloader
+and firmware-update flows.
+
+This module is exactly that: a tiny UDP server that speaks Trezor wire
+**protocol v1** (plaintext) on the emulator port. The bootloader always uses
+protocol v1 on every model (THP is firmware-only), so a single v1 mock covers
+both T2T1 and T3W1. The bridge discovers it over UDP just like a real emulator.
+
+It answers the handful of messages a host sends in bootloader mode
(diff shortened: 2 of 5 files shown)


OriginalEveres:
Validated end-to-end: with the mock running on the emulator UDP port and the node bridge up, Suite reports **"Device connected in bootloader mode"** (T3W1). The bridge `/call` `BridgeProtocolMessage` format that couldn't be driven from the Python test harness is produced natively by Suite/@trezor-connect, so the full path works.

OriginalEveres:
Button wired up and verified in a rebuilt image: dashboard **Bootloader mode (mock) → Start bootloader mock** sends `emulator-start-bootloader` → mock comes up on the emulator port and the node bridge enumerates it; **Stop** tears it down (enumerate goes empty). Also fixed: the start response no longer sets `emulator_started`, so the dashboard doesn't try to show a (nonexistent) VNC viewer for the mock.

Lemonexe:
Hmm, interesting :thinking:
**TL;DR:** IMO better not to have anything than a potentially misleading mock that leads you astray.

It is occasionally desired to test devices in bootloader, and currently you have to use physical devices which is less practical, but IDK if this is the right approach..
- It's unprecedented to emulate device directly in user env code.
- As I understand it, user-env is supposed to emulate the device as closely to production behavior, and that's why it runs a real FW build.
  - Even currently there are many differences between user env and real device, but the point is, that it should follow as closely as possible – and cover only those features where that's feasible.
- That's why user env is used for E2E as a fairly reliable API contract, but this PR is only a mock.
- → I'm afraid that if we have this in user env, it will lead devs to consider this simple mock behavior as an API contract, but it's more like a mock for unit tests, can be used for testing something very specific, but not a "general simulation". It hides many behaviors and potential anomalies.

:thought_balloon: alternative: how about doing it in Suite instead? That would make it clear we're not simulating a device, just mocking it. Possibly could implemented as a Transport, turned on with env variable, so that Connect uses this transport which will present a mock device to Connect. Kinda like the reverse of `@trezor/connect-cli`

szymonlesisz:
there is a `bootloader_emu` in trezor-firmware repo and it can be started in userenv instead of a mock:

https://github.com/trezor/trezor-firmware/blob/main/.github/actions/build-core-emu/action.yml#L30

PR #404, opened 2026-08-18
feat: tropic emulator loader

szymonlesisz:
Using the code from [tezor-firmware ](https://github.com/trezor/trezor-firmware/blob/main/python/src/trezorlib/_internal/emulator.py#L229) repo to prevent race conditions when tropic emulator is starting


mmilata (approved):
Untested but LGTM

